PR #49, opened 2019-06-11
Fix build on mac and add WeightNorm.module() getter

lunixbochs:
I'm trying to export weights from my wav2letter models and WeightNorm didn't have an obvious way of getting at the underlying Conv2d module.


lunixbochs:
What's this waiting on? I basically need to maintain a flashlight fork to be able to export wav2letter modules until the module() getter is merged, which makes it harder for people to build my wav2letter fork.

jacobkahn:
Apologies for the delay @lunixbochs — importing. Looks fine to me.

PR #72, opened 2019-10-30
Automatic Mixed Precision (AMP) support in flashlight

mtmd:
Summary:
Implements Automatic Mixed Precision (AMP) support in Flashlight. All appropriate modules will support half precision for their computations and use mixed precision (a combination of half precision and full precision) to strike a balance between accuracy and speedup. It also enables Tensor Cores when applicable to further accelerate the training.


flashlight/autograd/Variable.h
```diff
@@ -116,6 +117,43 @@ class Variable {
    */
   af::array& array() const;
 
+  /**
+   * Casts the variable into an appriperiate type based on the value of
```
WilliamTambellini: appriperiate ?

flashlight/meter/PerformanceAudit.h
```diff
@@ -0,0 +1,347 @@
...
+
+/**
+ * Determines the math type that should be used for a certain CUDA kernel
+ * (calculating a convolution in both forward and backward may incur multiple
+ * CUDA kernel calls).
+ * F32: Computation must be performed in F32.
+ * F32_ALLOW_CONVERSION: Input and output are in F32, however, we give
+ * permission to the algorithm to convert to F16 if it determines such
+ * conversion is beneficial.
+ * F16: Input and output are in F16. Hence, not internal conversion will occur.
```
WilliamTambellini: no internal ?

flashlight/autograd/Variable.h
mtmd: Thank you, @WilliamTambellini. Fixed. Will be part of the next commit.

flashlight/meter/PerformanceAudit.h
mtmd: Thank you, @WilliamTambellini. Fixed. Will be part of the next commit.

examples/DistributedTraining.cpp
```diff
@@ -30,7 +30,8 @@ int main() {
   auto reducer = std::make_shared<fl::CoalescingReducer>(
       /*scale=*/1.0 / worldSize,
       /*async=*/true,
-      /*contiguous=*/true);
+      /*contiguous=*/true,
+      /*useMixedPrecision*/ false);
```
jacobkahn: Instead of passing this bool flag or even an `af::dtype`, let's do what I suggested above for autograd and Modules - let's look at the `af::dtype` if the `af::array` which we're synchronizing and choose which NCCL param type we use based on that so there aren't any unexpected casts.

flashlight/autograd/Functions.cpp
```diff
@@ -50,6 +50,7 @@ Variable operator+(const Variable& lhs, const Variable& rhs) {
   auto result = lhs.array() + rhs.array();
   auto gradFunc = [](std::vector<Variable>& inputs,
                      const Variable& grad_output) {
+    typeTrace("Operator+ BWD", grad_output.type());
```
jacobkahn: Just curious - I understand these are probably helpful for debugging; is the plan to remove these before merging?

flashlight/autograd/Functions.cpp
```diff
@@ -236,35 +246,51 @@ Variable operator!(const Variable& input) {
   return Variable(result, false);
 }
 
-Variable max(const Variable& lhs, const Variable& rhs) {
+Variable
+max(const Variable& lhs, const Variable& rhs, const bool useMixedPrecision) {
```
jacobkahn: The nice thing about removing this flag and removing `autoCast` is that we don't need any autocast logic at all - we basically just do the operation in whatever precision the input Variables had, which ArrayFire will do by default.

flashlight/autograd/backend/cpu/Conv2D.cpp
```diff
@@ -42,9 +42,26 @@ Variable conv2d(
     int py,
     int dx,
     int dy,
-    int groups) {
+    int groups,
+    bool useMixedPrecision,
+    const ConvPerfAuditPtr audit) {
+  if (useMixedPrecision) {
+    throw std::runtime_error("Half precision is not supported in CPU.");
```
jacobkahn: We can handle this case similarly by checking the `input.dtype()` and throwing if it `== af::dtype::f16`.

flashlight/autograd/backend/cpu/Conv2D.cpp
```diff
@@ -42,9 +42,26 @@ Variable conv2d(
     int py,
     int dx,
     int dy,
-    int groups) {
+    int groups,
+    bool useMixedPrecision,
+    const ConvPerfAuditPtr audit) {
```
jacobkahn: Is the `ConvPerfAuditPtr` also for debugging benchmarking? I'd prefer if we removed it before merging.

flashlight/autograd/backend/cuda/BatchNorm.cpp
```diff
@@ -131,70 +136,80 @@ Variable batchnorm(
...
+      DevicePtr grad_in_raw(grad_in.array());
+      DevicePtr grad_wt_raw(grad_wt.array());
+      DevicePtr grad_bs_raw(grad_bs.array());
+
+      DevicePtr grad_op_raw(grad_output.array());
+
+      DevicePtr save_mean_raw(save_mean);
+      DevicePtr save_var_raw(save_var);
+
+      CUDNN_CHECK_ERR(cudnnBatchNormalizationBackward(
```
jacobkahn: Noticing that this logic hasn't changed at all for the AMP case - is this intentional?

flashlight/autograd/backend/cuda/Conv2D.cpp
```diff
@@ -6,17 +6,49 @@
...
+    CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM,
+    CUDNN_CONVOLUTION_FWD_ALGO_WINOGRAD_NONFUSED};
+
+const u_set<cudnnConvolutionBwdDataAlgo_t> bwdDataPreferredAlgos = {
+    CUDNN_CONVOLUTION_BWD_DATA_ALGO_1,
+    CUDNN_CONVOLUTION_BWD_DATA_ALGO_WINOGRAD_NONFUSED};
+
+const u_set<cudnnConvolutionBwdFilterAlgo_t> bwdFilterPreferredAlgos = {
+    CUDNN_CONVOLUTION_BWD_FILTER_ALGO_1,
+    CUDNN_CONVOLUTION_BWD_FILTER_ALGO_WINOGRAD_NONFUSED};
```
jacobkahn: Good to know — thanks for adding this! 👍

flashlight/autograd/backend/cuda/Conv2D.cpp
```diff
@@ -25,25 +57,70 @@ constexpr cudnnConvolutionBwdDataAlgo_t kBwdDataDefaultAlgo =
...
 // Get the algorithm which gives best performance.
 // Since cuDNN doesn't support memory limits, we manually choose an algorithm
 // which requires less than a specific workspace size.
-template <typename T>
-T getBestAlgorithm(const std::vector<T>& algoPerfs) {
+template <typename T, typename ALGO_TYPE>
+T getBestAlgorithm(
+    const std::vector<T>& algoPerfs,
+    const u_set<ALGO_TYPE>& preferredAlgos,
+    const bool algoMustSupportF16) {
```
jacobkahn: As above, it's best if bool flags like this were passed as `af::dtype`s directly so that we can easily add support for other precisions and keep everything less opinionated.

flashlight/meter/PerformanceAudit.h
```diff
@@ -0,0 +1,347 @@
+/*
```
jacobkahn: As above - you're planning on removing this after you're done benchmarking, yes?

flashlight/distributed/reducers/CoalescingReducer.h
```diff
@@ -26,6 +26,7 @@ class Variable;
  * ``finalize`` must be called before using a given value.
  */
 class CoalescingReducer : public Reducer {
+  bool useMixedPrecision_;
```
jacobkahn: See my above comment. We should just dynamically choose the `ncclType` based on the `af::dtype` if the Variable to be reduced. That way we don't potentially accidentally lose precision in the reduction if we sync parameters from an fp32 module.

flashlight/optim/Optimizers.cpp
```diff
@@ -168,8 +175,24 @@ void AdamOptimizer::step() {
...
+      af::eval(dataF32);
+      data = dataF32.as(f16);
+    } else {
+      if (wd_ != 0) {
+        // Weight decay term
+        data = data - wd_ * data;
+      }
+      data =
+          data - (correctedLr * biasedFirst) / (af::sqrt(biasedSecond) + eps_);
+    }
```
jacobkahn: We want to do all parameter and gradient accumulations in fp32 for now. There's a fair amount of research supporting that this is important in seq2seq domains (happy to share); there shouldn't be a need to check if the accumulator is fp16.

flashlight/nn/modules/Module.h
```diff
@@ -44,6 +44,17 @@ class Module {
...
+   * A golden reference of parameters (FP32) when mixed precision mode is
+   * enabled.
+   */
+  std::vector<af::array> paramsF32_;
+
+  /**
+   * Determines if the module is allowed to uses mixed precision.
+   */
+  bool useMixedPrecision_ = false;
+
```
jacobkahn: As above, we don't need to serialize the mixed precision flag and can always serialize the fp32 params for now. We can add functionality to change the precision of the weights we serialize later.

mtmd:
Thank you very much @jacobkahn for the detailed feedback :). I have a few questions, let's discuss them one at a time to ensure there will be no misunderstanding. I had the same feeling about the `useMixedPrecision` flag. It's possible to move forward with it's removal in modules that don't have parameters, and achieve the same functionality with the cast that you suggested. However, when a module has parameters (e.g., convolution) I use the flag to create a master copy (single precision) of parameters on instantiation. Subsequently, in each iteration, the master copy gets updated during the optimization and the half precision copy will be used for computation.  How are we planing to take care of this after we remove the `useMixedPrecision` flag?

flashlight/autograd/Functions.cpp
mtmd: As you mentioned, it is useful for debugging. In particular, if in future a module gets revised and starts generating output in an undesired data type (e.g., due to a type promotion), this function can be used to quickly detect the reason for the performance hit. Otherwise, the program might be casting large tensors back and forth without us noticing. In terms of performance, It doesn't adversely impact the execution time, since it gets completely optimized away when the `TRACE_PRECISION` is not defined.

flashlight/autograd/backend/cpu/Conv2D.cpp
mtmd: `ConvPerfAuditPtr` is not for debugging. It helps with choosing the convolution algorithm that yields the highest performance. Depending on the convolution kernel, there are cases that performing the convolution in F32 can actually yield a better performance (since not all algorithms can support half precision). In those cases, even if the data is in F16, it is beneficial to perform the convolution in F32. I understand that it adds complication, so we can remove it. However, if we remove it, there will be significant performance loss on some networks.

flashlight/autograd/backend/cuda/BatchNorm.cpp
mtmd: Actually for BN, if I understand your question correctly, providing input tensors with correct type is all we need to do. In the current version, we take care of that with the `autoCast`.

flashlight/optim/Optimizers.cpp
mtmd: @jacobkahn sure. Can you please share some of those articles with me to take a look at them?

jacobkahn: https://arxiv.org/pdf/1710.03740.pdf is the canonical one, but there's also https://arxiv.org/pdf/1802.00930.pdf and https://arxiv.org/pdf/1807.11205.pdf which discuss the update a bit.

jacobkahn:
@mtmd to answer your questions:

>  use the flag to create a master copy (single precision) of parameters on instantiation

You're right - with the implementation without the flag, we always store parameters in FP32. The API we came up with with respect to this is similar to what you have, except we add a constructor that also allows an `af::dtype` to be passed, which specifies how the parameters are actually stored. That way, we can still use an `fl::Variable` and `af::array`, but will just cast the parameters upon instantiation.

That's a separate change to the other operators though, so I think it makes sense to implement that either in a separate PR or a separate commit, if that makes sense. Preliminarily, params in modules can always be fp32.

On artifact or always keeping fp32 master and converting to fp16 on each iteration is that we'll have to convert then on the fly, as you say. With ArrayFire, since there are no in-place operations anyways with the JIT, converting them on the fly is just as good as keeping a separate copy, since that underlying memory can change. This is something that the memory manager can take care of, but we don't need to store an fp16 array on the module, since it might go unused and won't save memory overhead.

> the master copy gets updated during the optimization and the half precision copy will be used for computation. How are we planing to take care of this

The simple solution is to check the input - if it's fp16, then convert the weights to fp16 on the fly and do the fp16 operation. Otherwise, use full precision. As I mentioned above, we can change down the line how the weights are stored, but we should have fp32 as the default option either way.

flashlight/autograd/backend/cuda/BatchNorm.cpp
jacobkahn: Ok. Once we remove `autoCast`, the tensor `DevicePtr` should already have the correct precision.

flashlight/autograd/backend/cpu/Conv2D.cpp
jacobkahn: Got it — that's helpful. Didn't quite understand what was happening there.

My primary question is — why do we actually need this? Why is the choice of algorithm empirical like this — can't cuDNN select this ahead of time based on heuristics similar to these and the cudnn math mode?

We can definitely have some sort of stateful thing which keeps tallies of performance for the operator — need to think about the implementation a bit.

mtmd:
@jacobkahn, Awesome! Will restart working on this repository late this week/early next week.

jacobkahn:
Hey @mtmd — how's it coming along? Do you have any questions?

flashlight/autograd/backend/cpu/Conv2D.cpp
jacobkahn: Also want to add — because the shape of the convolutions that we perform changes so frequently, we never use `cuDNN` benchmark mode, so those heuristics don't really do much for us. I'd be interested in seeing a benchmark with and without the Conv audit.

mtmd:
> Hey @mtmd — how's it coming along? Do you have any questions?

Hi Jacob (@jacobkahn), not at the moment, thank you for asking :). Prepared and pushed a new commit. Should include almost everything that we have discussed so far.

flashlight/nn/modules/Module.cpp
```diff
@@ -88,6 +90,25 @@ Variable UnaryModule::operator()(const Variable& input) {
   return this->forward(input);
 }
 
+PrecisionCast::PrecisionCast(af::dtype targetType) : targetType_(targetType) {}
```
jacobkahn: Can we move `PrecisionCast` to its own file rather than defining it in `Module`?

flashlight/nn/modules/Module.h
```diff
@@ -163,6 +169,23 @@ class UnaryModule : public Module {
   FL_SAVE_LOAD_WITH_BASE(Module)
 };
 
+/**
+ * Precision Cast Module. Casts the input from its original precision to the
+ * target precision. Precision cast only alters the underlying array and leaves
+ * other attributes of the input variable unchanged.
+ */
+
+class PrecisionCast : public UnaryModule {
```
jacobkahn: Why is precision cast a unary module? Shouldn't it be able to operate on a vector of `fl::Variable` as well and cast each element?

flashlight/meter/PerformanceAudit.h
```diff
@@ -0,0 +1,353 @@
...
+    bwdFilter->setUseMixedPrecision(useMixedPrecision);
+  }
+};
+
+using ConvPerfAuditPtr = std::shared_ptr<ConvPerfAudit>;
+
+/**
+ * @return returns a pointer to a ConvPerfAudit struct.
+ */
+inline ConvPerfAuditPtr getConvAuditHandle() {
```
jacobkahn: Still not crazy about the perf audit, but starting to warm up to it a bit. Going to give more thought to how we might want to add this and will give feedback shortly afterwards.

flashlight/autograd/Functions.h
```diff
@@ -361,6 +363,18 @@ split(const Variable& input, const std::vector<dim_t>& splitSizes, int dim);
...
+ * Repeats the tensor `input` along specific dimensions. The number of
+ * repetition along each dimension is specified in descriptor `dims`.
+ *
+ * @param[in] precision Type of the output vector when is it is desired to
+ * be different from the input type. This is particularly useful when tile is
+ * applied on parameters and the results will be used in a half precision
+ * arithmetic.
+ */
+Variable
+tile(const Variable& input, const af::dim4& dims, const af::dtype precision);
```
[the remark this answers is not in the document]
mtmd: Yes, sometimes we use this function to tile the weights. In such settings, it would be beneficial to use the target type instead of weight type (i.e., f32). Please see line: 41 in  LayerNorm.cpp.
The function is overloaded. The default one (line 364 in Functions.h) inspects the input type and performs the tilling according to that.

flashlight/autograd/backend/cuda/BatchNorm.cpp
```diff
@@ -66,26 +69,44 @@ Variable batchnorm(
...
-  auto weight_nonempty = weight.isempty()
+  auto weightNonempty = weight.isempty()
       ? Variable(af::constant(1.0, wt_desc_dims, input.type()), false)
       : weight;
-  auto bias_nonempty = bias.isempty()
+  auto biasNonempty = bias.isempty()
       ? Variable(af::constant(0.0, wt_desc_dims, input.type()), false)
       : bias;
 
+  af::array* weightArrayPtr;
```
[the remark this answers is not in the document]
mtmd: Yes, that was the concern. Sure. Moved ahead and replaced all the pointers.

flashlight/autograd/backend/cuda/BatchNorm.cpp
```diff
@@ -145,25 +168,46 @@ Variable batchnorm(
             : inputs[1];
         auto& bs = inputs[2];
 
-        const void* one1 = kOne(in.type());
-        const void* zero0 = kZero(in.type());
+        af::array* grad_output_array_ptr;
```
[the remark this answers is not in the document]
mtmd: Done.

flashlight/autograd/backend/cuda/RNN.cpp
```diff
@@ -39,11 +38,42 @@ void rnn_backward(
...
+  typeTrace("RNN BWD - hx", inputs[1].type());
+  typeTrace("RNN BWD - cx", inputs[2].type());
+
   auto input = inputs[0];
   auto hx = inputs[1];
   auto cx = inputs[2];
   auto weights = inputs[3];
 
+  af::array hxArray;
+  af::array* hxArrayPtr;
```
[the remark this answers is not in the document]
mtmd: Done.

flashlight/autograd/backend/cuda/RNN.cpp
```diff
@@ -62,6 +92,10 @@ void rnn_backward(
   DropoutDescriptor dropout(drop_prob);
   RNNDescriptor rnn_desc(
       input.type(), hidden_size, num_layers, mode, bidirectional, dropout);
+  if (input.type() == f16) {
```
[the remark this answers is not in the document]
mtmd: Yes, it should handle all cases. Note that we reinitialize the descriptor in every iteration. As a result, it uses `f32` except if we instruct it otherwise. Nonetheless, it's good to add a case for `f32` to increase the readability. I went ahead and revised the code accordingly.

flashlight/common/Defines.h
```diff
@@ -10,6 +10,8 @@
 
 #include <cstdlib>
 
+#define __forceinline __attribute__((always_inline)) inline
```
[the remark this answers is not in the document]
mtmd: No. Choice of always inline was made to ensure the critical code section would not include overhead of function call in case `-fno-inline` is set or the inlining limit is reached. The former is the default when not optimizing ([more](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html)).

flashlight/common/Utils.cpp
```diff
@@ -26,4 +28,48 @@ bool allClose(
...
+      {u64, "u64"},
+      {s16, "s16"},
+      {u16, "u16"},
+      {f16, "f16"},
+  };
+  if (typeToStr.find(type) != typeToStr.end()) {
+    return typeToStr[type];
+  }
+  throw std::invalid_argument("Invalid inputs type: " + type);
+}
```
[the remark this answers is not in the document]
mtmd: Yes:
`stringToAfType()` is used in wav2letter when casting the input arch file.
`afTypeToString()` is used in `prettyString()` method of `PrecisionCast` class.

flashlight/common/Utils.h
```diff
@@ -8,21 +8,59 @@
...
+    return static_cast<std::size_t>(t);
+  }
+};
+
+template <typename T>
+using u_set = std::unordered_set<T, EnumNaiveHash>;
+
+template <typename KeyType, typename ValueType>
+using u_map = std::unordered_map<KeyType, ValueType, EnumNaiveHash>;
+
```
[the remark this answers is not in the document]
mtmd: That's correct. However, `u_set` is used in `Conv2D.cpp`.

flashlight/distributed/reducers/CoalescingReducer.cpp
```diff
@@ -22,6 +22,7 @@ CoalescingReducer::~CoalescingReducer() {
 }
 
 void CoalescingReducer::add(Variable& var) {
+  var.cast(f32);
```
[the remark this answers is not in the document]
mtmd: Precisely. However, there is a bit complication involved. That's why I initially moved ahead with a naive implementation: 
1. In a lazy approach, a cast to a higher precision will increase the current size of `cache_` by ~2X. The new size can be larger than the the constant threshold that is set for flushing the data (i.e., `kCoalesceCacheSize`). This is gonna break other assumptions regarding when we actually flush, since flushing can happen anywhere in the range of `[kCoalesceCacheSize, 2*kCoalesceCacheSize)`.
2. We can find a remedy for the aforementioned issue, but also we need to reconsider the choices that are made for coalesced memory allocation in `NcclContext::getCoalesceBuffer()`.

Perhaps we can keep this in the naive form for this PR in favor of agility.

flashlight/nn/modules/Module.cpp
mtmd: Done.

flashlight/nn/modules/Module.h
mtmd: True. Fixed.

flashlight/meter/PerformanceAudit.h
mtmd: Sure. Feel free to share your thoughts on how you'd prefer to change it :).

jacobkahn: See my comment below.

jacobkahn:
@mtmd — after discussing further with the team, we've decided we want to remove all of the perf audit stuff for now. We know not having it incurs a performance penalty, but we need to take some time to think about how the implementation interacts with the overall framework and how it can be made generic enough. We'd still like to merge the other parts of the change and we're really excited about them!

Once you remove that, pending a few other of my comments above, we should be ready to squash and merge everything. Thanks again for the outstanding work and patience here!

flashlight/distributed/reducers/CoalescingReducer.cpp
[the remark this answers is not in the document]
jacobkahn: These are great points. I do have ways to resolve both of these issues, but you're right that for now, we can move forward with a naive implementation (Y) and make those other changes down the line.

mtmd:
> @mtmd — after discussing further with the team, we've decided we want to remove all of the perf audit stuff for now. We know not having it incurs a performance penalty, but we need to take some time to think about how the implementation interacts with the overall framework and how it can be made generic enough. We'd still like to merge the other parts of the change and we're really excited about them!
> 
> Once you remove that, we should be ready to squash and merge everything. Thanks again for the outstanding work and patience here!

@jacobkahn Thank you, Jacob. Removed it in 5dd5ea5.

mtmd:
Hey Jacob, I went over our conversations and it seems like there are no more pending action items. If correct, please let me know to move forward with squashing.

jacobkahn:
@mtmd - yes, everything looks good. A few other notes as I start merging things in.
- When I merge the PR, it'll squash all the commits automatically - you don't have to do any squashing on your side.
- For now, we're going to merge this into separate branches inside flashlight and wav2letter. Because this change doesn't compile if you're not using ArrayFire master and the 3.7 release still won't be ready for a few more weeks, we want to make sure that master still builds with stable ArrayFire releases. We'll use those branches in flashlight and wav2letter to continue work on this as needed; once AF 3.7 is available, we'll send another PR to merge those into master.

mtmd:
@jacobkahn Sounds great. Please feel free to let me know if we need more work on this thread. Also, please let me know when we want to start reviewing the corresponding PR in w2l repository.

jacobkahn:
@mtmd — I've started reviewing https://github.com/facebookresearch/wav2letter/pull/435 as well — I hope to finish in the next day or two. I'm actually now leaning towards waiting until AF 3.7 is done and merging this directly into master.

mtmd:
Thank you @jacobkahn. Will Take care of it next week.

flashlight/nn/modules/PrecisionCast.cpp
```diff
@@ -0,0 +1,61 @@
...
+namespace fl {
+
+PrecisionCast::PrecisionCast(af::dtype targetType) : targetType_(targetType) {}
+
+std::vector<Variable> PrecisionCast::forward(
+    const std::vector<Variable>& inputs) {
+  std::vector<Variable> outputs;
+  for (auto input : inputs) {
+    auto output = input.withoutData();
+    output.array() = input.array().as(targetType_);
```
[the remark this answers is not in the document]
mtmd: Changed.

flashlight/autograd/Variable.h
```diff
@@ -116,6 +117,18 @@ class Variable {
...
+  /**
+   * Casts the variable to the given data type.
+   *
+   * @param[in] type target data type
+   */
+  __forceinline void cast(af::dtype type) {
+    if (this->type() == type) {
+      return;
+    }
+    array() = array().as(type);
```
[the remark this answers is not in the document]
mtmd: Fixed.

mtmd:
> @mtmd — the wait's over - flashlight is finally bumped to ArrayFire 3.7.1, so this PR should build with everything else! Would be great if you could rebase your fork on top of flashlight master.
> 
> I also have very old draft review comments below - I'm not sure if I already submitted them or if they're relevant anymore, but submitting nonetheless.

Thank you @jacobkahn for your comments. Rebased, addressed the issues, and pushed.

jacobkahn:
@mtmd — just rebased the `amp` branch, on master, so if you could rebase once more, that would be great.

flashlight/autograd/Variable.h
```diff
@@ -116,6 +117,42 @@ class Variable {
...
   af::array& array() const;
 
+  /**
+   * Casts the variable to the given data type.
+   *
+   * @param[in] type target data type
+   * @param[in] castGrad determines if the gradient should be casted in case it
+   * exists.
+   */
+  __forceinline void inPlaceCast(af::dtype type, bool castGrad = true) {
```
jacobkahn, line 127: Is there some reason to define these in the header/inline? Would prefer not to.

jacobkahn: Also, I'm not sure if a situation in which we'd want `castGrad` to *not* be true. If that were the case, we could just potentially set the gradient ourselves, but it seems like an edge case. I think we should remove the argument/condition.

mtmd: Done, done, rebased, and pushed.

flashlight/nn/modules/PrecisionCast.cpp
```diff
@@ -0,0 +1,60 @@
+/*
```
jacobkahn, line 1: Please remove the copyright header and license from files that have them. Is there a reason you've added them?

flashlight/nn/modules/PrecisionCast.h
```diff
@@ -0,0 +1,56 @@
+/*
```
jacobkahn, line 1: ditto with copyright/license header.

mtmd:
> @mtmd — can you rebase again? For whatever reason, it seems there are still a lot of changes from master that are showing up as being tracked in the PR. I can also rebase your fork if you're having trouble. `flashlight:amp` should be even with master as before.
> 
> Also, see my below comments.

Oh! Just noticed. Done.

flashlight/nn/modules/PrecisionCast.cpp
mtmd: >> workplace.

flashlight/nn/modules/PrecisionCast.h
mtmd: => workplace.

flashlight/nn/Utils.h
```diff
@@ -13,13 +13,38 @@
...
 
+#include <iomanip>
+
 #include <af/dim4.hpp>
 
 #include "flashlight/common/Defines.h"
+#include "flashlight/common/Utils.h"
 #include "flashlight/nn/modules/Module.h"
 
+#ifdef TRACE_PRECISION
```
jacobkahn, line 24: Can we put this in an anonymous namespace?

flashlight/nn/modules/Conv2D.h
```diff
@@ -8,9 +8,11 @@
 
 #pragma once
 
+#include <memory.h>
```
jacobkahn, line 11: remove? I assume this is something internal :)

flashlight/nn/modules/PrecisionCast.h
```diff
@@ -0,0 +1,28 @@
...
+ * Precision Cast Module. Casts the input from its original precision to the
+ * target precision. Precision cast only alters the underlying array and leaves
+ * other attributes of the input variable unchanged.
+ */
+class PrecisionCast : public Module {
+  af::dtype targetType_;
+  PrecisionCast() = default;
+  FL_SAVE_LOAD_WITH_BASE(Module, targetType_)
+ public:
+  PrecisionCast(af::dtype targetType);
```
jacobkahn, line 19: I know they're trivial, but can we add doxygen docs for these two?

Would be nice to be precise about the behavior that `std::vector<Variable> forward(const std::vector<Variable>& inputs) override;` casts every input.

flashlight/nn/modules/PrecisionCast.h
```diff
@@ -0,0 +1,28 @@
...
+  std::vector<Variable> forward(const std::vector<Variable>& inputs) override;
+  Variable forward(const Variable& input);
+  Variable operator()(const Variable& input);
+  std::string prettyString() const override;
+};
+
+} // namespace fl
+
+CEREAL_REGISTER_TYPE(fl::PrecisionCast)
```
jacobkahn, line 28: newline at end of file please

flashlight/nn/Utils.h
mtmd: Sure.

flashlight/nn/modules/Conv2D.h
mtmd: Done.

flashlight/nn/modules/PrecisionCast.h
mtmd: Sure.

flashlight/nn/modules/PrecisionCast.h
mtmd: Done.

PR #84, opened 2020-01-21
[Feature Addition] Mish activation

thebhatman:
I have added the Mish Activation function. This resolves  https://github.com/facebookresearch/flashlight/issues/81. I am assuming the autograd would take care of the gradients. If I'm wrong, I'll add the gradient definition as well.

### DIFF
--- a/docs/source/modules.rst
+++ b/docs/source/modules.rst
@@ -41,6 +41,9 @@ Activations
 .. doxygenclass:: fl::Sigmoid
   :members:
 
+.. doxygenclass:: fl::Mish
+  :members:
+
 .. doxygenclass:: fl::Tanh
   :members:
 
--- a/flashlight/nn/modules/Activations.cpp
+++ b/flashlight/nn/modules/Activations.cpp
@@ -31,6 +31,16 @@ std::string Sigmoid::prettyString() const {
   return "Sigmoid";
 }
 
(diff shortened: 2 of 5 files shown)


avidov:
Thank you for the contribution.

> I am assuming the autograd would take care of the gradients.
Will you please add an autograd test in flashlight/tests/autograd/AutogradTest.cpp ?

Also please add a simple serialization test tests/nn/SerializationTest.cpp

avidov:
Thank you for adding the serialization test.
My other request, in the comment above, was grayed out so I'll repeat it here.

Please add an autograd test in flashlight/tests/autograd/AutogradTest.cpp ?

thebhatman:
I have added the test for Autograd. Can you please review it and check if it's the right way to implement? I am comparing the gradient obtained by autograd with the actual differentiation of mish.

jacobkahn:
@thebhatman — we appreciate the contribution. After some discussion, we don't think this should be included in flashlight (at this moment).

We strive to include only methods, operators, and modules that have proven long-term success and results. This keeps our codebase lightweight, reduces the amount of maintenance we have to perform, and ensures a lack of opinionation.

If, down the line, the activation does show robustness across a variety of domains and tasks, we can certainly consider including it! Thanks again for the contribution.

PR #91, opened 2020-03-07
Fix wrong file name in tests/CMakeLists.txt

hanbyul-kim:
Build error has happened due to wrong cpp file name. The name of the file, memory/ConfigurableMemoryAllocator.cpp, was changed but it wasn't reflected on the CMakeLists.txt.


jacobkahn:
Thanks for flagging @hanbyul-kim — previously fixed in 6170babdc5995fdedb98abf5dcc70c33961427f6 👍  - closing.

PR #129, opened 2020-05-22
Caching mem mgr oom

WilliamTambellini:


tests/memory/CachingMemoryManagerTest.cpp
```diff
@@ -120,6 +120,27 @@ TEST_F(CachingMemoryManagerTest, LargeNumberOfAllocs) {
   }
 }
 
+TEST_F(CachingMemoryManagerTest, OOM) {
+  af_backend b;
+  af_get_active_backend(&b);
+  // Despite that test is trying to allocate PB of memory,
+  // it could theoretically pass with some CPU backends (afcpu or afopencl using intel cpu drivers) when doing lazy mem allocation.
```
vineelpratap, line 127: May be better to use `af::randu`. It is not evaluated lazily ?

WilliamTambellini: sure sure done

PR #181, opened 2020-09-24
Adding max & min functions along given dim

wesbz:
### Summary
This pr adds max & min functions along given dim of a fl::Variable.

### Test Plan (required)
Created a test function in AutogradTest.cpp which yielded expected results.

### DIFF
--- a/flashlight/autograd/Functions.cpp
+++ b/flashlight/autograd/Functions.cpp
@@ -247,6 +247,18 @@ Variable operator!(const Variable& input) {
   return Variable(result, false);
 }
 
+Variable max(const Variable& input, const int& dim) {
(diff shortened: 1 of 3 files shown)


flashlight/autograd/Functions.cpp
```diff
@@ -272,6 +284,18 @@ Variable max(const double& lhsVal, const Variable& rhs) {
   return max(rhs, lhsVal);
 }
 
+Variable min(const Variable& input, const int& dim) {
+  auto result = min(input.array(), dim);
+  auto gradFunc = [result, dim](std::vector<Variable>& inputs,
```
WilliamTambellini, line 289: Hum, some code duplication here : the only diff seems to be the result array, right ?

PR #188, opened 2020-09-29
Enhance the CachingMemoryManager to possibly handle high memory pressure

### LINKED ISSUE #180: Enhance the CachingMemoryManager for low mem usecase
### Feature Description
As today the CachingMemoryManager is faster than the default AF mem mgr but greedy in memory preventing to train big models on low memory GPUs. This feature proposes to add 2 new params to the CachingMemMgr in order to bypass this limitation. 

#### Use Case
Scenario showing the limit of the current CachingMemMgr, considering a 12GB GPU:
- allocate an array of 10GB
- release the variable/AF array : the 10GB mem buf is not free but just registered to the large free pool ("largeBlocks_") for later recycling
- create a 5GB array : see the CachingMemMgr recycles the previous 10GB buffer by splitting the block in 2 creating a free block of 5GB for the remaining right side of the buffer:
https://github.com/facebookresearch/flashlight/blob/e692a941e902f9e207a1a4ee226c0ad0f8dbdc5e/flashlight/memory/managers/CachingMemoryManager.cpp#L149
- try to create a 6GB array : the available 5GB block is too small and cannot be free because being a 'split', the GPU still has 2GB of unallocated mem, meaning the GPU still has total 7G of unused mem (5+2), but see that the CachingMemMgr OOM. 
This proposal would allow to bypass this limitation. 

[Example usage of the feature - why is it needed?]
It is needed in order to train bigger model without the need for a high mem capacity GPU.
I have found at least 1 way to enhance the CachingMemMgr to be able to handle such scenario by just adding 2 new params to be set accordingly by the user, the default value not changing the current behavior :
size_t recyclingSizeLimit = std::numeric_limit<size_t>::max(); // prevent to recycle buffers, to be set by the user if desired
size_t splitSizeLimit = std::numeric_limit<size_t>::max(); // prevents to split big buffers, to be set by the user if desired
to be used in CachingMemoryManager::alloc(...):
https://github.com/facebookresearch/flashlight/blob/e692a941e902f9e207a1a4ee226c0ad0f8dbdc5e/flashlight/memory/managers/CachingMemoryManager.cpp#L111

1 in order to control the recycling gate : 
https://github.com/facebookresearch/flashlight/blob/e692a941e902f9e207a1a4ee226c0ad0f8dbdc5e/flashlight/memory/managers/CachingMemoryManager.cpp#L130
 if (it != pool.end() && ( isSmallAlloc || (*it)->size_ < recyclingSizeLimit)) {

2 in order to prevent the split of big buffers generating fragmentation:
https://github.com/facebookresearch/flashlight/blob/e692a941e902f9e207a1a4ee226c0ad0f8dbdc5e/flashlight/memory/managers/CachingMemoryManager.cpp#L148
 if ((block->size_ < splitSizeLimit) // dont split big buffers
   && (diff >= (isSmallAlloc ? kMinBlockSize : kSmallSize))) {

By adding these 2 new params and fine setting it accordingly (in my case, recyclingLimit = 400MB and splitLimit=2GB), I ve been able to train some big models on a 12GB GPU where the current CachingMemMgr without these 2 options was needing a 24GB GPU.
Reminder: the default values (max) dont change the current behavior.

Note:
For the cuda backend, a long term enhancement would be to take advantage of nvidia vmm in order to potentially defragment the buffers but it would mean many changes in perhaps both AF and probably FL: 
https://developer.nvidia.com/blog/introducing-low-level-gpu-virtual-memory-management/

Considering the default behavior would not change, would you accept a PR for me to add these 2 params?
Kind
W.

WilliamTambellini:
Enhance the CachingMemoryManager to handle high memory pressure by adding 2 options.
Note: the default behavior is unchanged.
Add a new API to MemoryManagerAdapter to setOption().
Side: remove some glog includes failing compilation if glog-dev not installed locally

**Original Issue**: 
https://github.com/facebookresearch/flashlight/issues/180
closes #180

### Summary
The CachingMemoryManager being greedy in memory. This PR adds 2 options to mitigate this.

### Test Plan (required)
Implemented a new unitest in order to show OOM with the default CachingMM and no OOM if the right option is set.

### DIFF
--- a/app/asr/CMakeLists.txt
+++ b/app/asr/CMakeLists.txt
@@ -87,9 +87,9 @@ add_executable(Train ${CMAKE_CURRENT_LIST_DIR}/Train.cpp)
 add_executable(Test ${CMAKE_CURRENT_LIST_DIR}/Test.cpp)
 add_executable(Decoder ${CMAKE_CURRENT_LIST_DIR}/Decode.cpp)
 
-target_link_libraries(Train flashlight-app-asr)
-target_link_libraries(Test flashlight-app-asr)
-target_link_libraries(Decoder flashlight-app-asr)
+target_link_libraries(Train flashlight-app-asr ${CMAKE_DL_LIBS})
+target_link_libraries(Test flashlight-app-asr ${CMAKE_DL_LIBS})
+target_link_libraries(Decoder flashlight-app-asr ${CMAKE_DL_LIBS})
 
 # --------------------------- Tests ---------------------------
 
--- a/flashlight/memory/managers/CachingMemoryManager.cpp
+++ b/flashlight/memory/managers/CachingMemoryManager.cpp
@@ -10,7 +10,9 @@
 
 #include <algorithm>
 #include <iostream>
+#include <limits>
 #include <mutex>
+#include <numeric>
 #include <stdexcept>
 #include <string>
 #include <vector>
@@ -87,6 +89,14 @@ CachingMemoryManager::CachingMemoryManager(
 
 void CachingMemoryManager::initialize() {}
 
+void CachingMemoryManager::setRecyclingSizeLimit(size_t limit) {
+  recyclingSizeLimit_ = limit;
+}
+
+void CachingMemoryManager::setSplitSizeLimit(size_t limit) {
+  splitSizeLimit_ = limit;
+}
+
 void CachingMemoryManager::shutdown() {
   signalMemoryCleanup();
 }
@@ -126,7 +136,9 @@ void* CachingMemoryManager::alloc(
 
   CachingMemoryManager::Block* block = nullptr;
   auto it = pool.lower_bound(&searchKey);
-  if (it != pool.end()) {
+  // Recycle blocks if any found, and if small alloc or the block size is not
+  // too large:
+  if (it != pool.end() && (isSmallAlloc || (*it)->size_ < recyclingSizeLimit_)) {
     block = *it;
     pool.erase(it);
     memoryInfo.stats_.cachedBytes_ -= block->size_;
@@ -144,7 +156,10 @@ void* CachingMemoryManager::alloc(
   // implementation simple.
   CachingMemoryManager::Block* remaining = nullptr;
   size_t diff = block->size_ - size;
-  if (diff >= (isSmallAlloc ? kMinBlockSize : kSmallSize)) {
+  if ((diff >= (isSmallAlloc ? kMinBlockSize : kSmallSize)) &&
+      (block->size_ < splitSizeLimit_) // possibly dont split large buffers to
+                                      // minimize risk of fragmentation
+  ) {
     remaining = block;
     block = new Block(size, block->ptr_);
     block->prev_ = remaining->prev_;
--- a/flashlight/memory/managers/CachingMemoryManager.h
+++ b/flashlight/memory/managers/CachingMemoryManager.h
@@ -7,8 +7,11 @@
 
 #pragma once
 
+#include <atomic>
+#include <limits>
 #include <memory>
 #include <mutex>
+#include <numeric>
 #include <set>
 #include <unordered_map>
 #include <vector>
@@ -48,6 +51,9 @@ class CachingMemoryManager : public MemoryManagerAdapter {
   bool jitTreeExceedsMemoryPressure(size_t bytes) override;
   void addMemoryManagement(int device) override;
   void removeMemoryManagement(int device) override;
+  // Set runtime options: RecyclingSizeLimit, SplitSizeLimit, ... Warning: not thread safe
+  void setRecyclingSizeLimit(size_t);
+  void setSplitSizeLimit(size_t);
 
   // Block denotes a single allocated unit of memory.
   struct Block {
@@ -133,6 +139,15 @@ class CachingMemoryManager : public MemoryManagerAdapter {
 
   void tryMergeBlocks(Block* dst, Block* src, BlockSet& freeBlocks);
   void freeBlock(Block* block);
+
+ private:
+  // Non-const runtime options in order to fine tune the behavior of this
+  // manager. Prevents to recycle some buffers, to be set by the user if
+  // desired:
+  size_t recyclingSizeLimit_{std::numeric_limits<size_t>::max()};
+  //size_t recyclingSizeLimit;
+  // Prevents to split big buffers, to be set by the user if desired:
+  size_t splitSizeLimit_{std::numeric_limits<size_t>::max()};
 };
 
 } // namespace fl
--- a/flashlight/test/memory/CachingMemoryManagerTest.cpp
+++ b/flashlight/test/memory/CachingMemoryManagerTest.cpp
@@ -20,9 +20,8 @@
 class CachingMemoryManagerTest : public ::testing::Test {
  protected:
   virtual void SetUp() override {
-    auto deviceInterface_ =
-        std::make_shared<fl::MemoryManagerDeviceInterface>();
-    auto adapter_ = std::make_shared<fl::CachingMemoryManager>(
+    deviceInterface_ = std::make_shared<fl::MemoryManagerDeviceInterface>();
+    adapter_ = std::make_shared<fl::CachingMemoryManager>(
         af::getDeviceCount(), deviceInterface_);
     installer_ = fl::cpp::make_unique<fl::MemoryManagerInstaller>(adapter_);
     installer_->setAsMemoryManager();
@@ -32,6 +31,8 @@ class CachingMemoryManagerTest : public ::testing::Test {
     af_unset_memory_manager();
   }
 
+  std::shared_ptr<fl::MemoryManagerDeviceInterface> deviceInterface_;
+  std::shared_ptr<fl::CachingMemoryManager> adapter_;
   std::unique_ptr<fl::MemoryManagerInstaller> installer_;
 };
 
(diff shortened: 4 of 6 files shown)


ext/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ include(${CMAKE_CURRENT_LIST_DIR}/common/CMakeLists.txt)
  
 # --------------------------- Tests ---------------------------
 
-if (FL_BUILD_TESTS)
+# The SeqBuilder test needs some contribs modules
+if (FL_BUILD_TESTS AND FL_BUILD_CONTRIB)
```
jacobkahn, line 11: Can we remove this and instead [gate the `SequentialBuilder` test](https://github.com/facebookresearch/flashlight/blob/master/ext/test/CMakeLists.txt#L6) to only build if `FL_BUILD_CONTRIB` is on? You can send a separate PR for this.

flashlight/memory/MemoryManagerAdapter.cpp
```diff
@@ -66,6 +66,10 @@ af_memory_manager MemoryManagerAdapter::getHandle() const {
   return interface_;
 }
 
+void MemoryManagerAdapter::setOption(unsigned key, const std::string& value) {
```
jacobkahn, line 69: I think we should pursue an approach where we put specific methods on memory manager implementations that support them rather than having string options. Doing things at compile time seems like it'll make things easier to reason about.

flashlight/memory/managers/CachingMemoryManager.cpp
```diff
@@ -88,6 +96,25 @@ CachingMemoryManager::CachingMemoryManager(
 
 void CachingMemoryManager::initialize() {}
 
+void safeStoull(size_t& var, const char* str) {
```
jacobkahn, line 99: Let's add this function in an anonymous namespace at the top of this file

flashlight/memory/managers/CachingMemoryManager.cpp
```diff
@@ -88,6 +96,25 @@ CachingMemoryManager::CachingMemoryManager(
...
+    return;
+  try {
+    auto value = std::stoull(str);
+    var = value;
+  } catch(const std::exception& e) {
+    // FL_LOG Warn ?
+  }
+}
+
+void CachingMemoryManager::setOption(unsigned key, const std::string& value) {
```
jacobkahn, line 110: As I mentioned, let's add setters on `CachingMemoryManager` directly for these parameters rather than using string option params.

jacobkahn (changes requested):
Nice. Thanks for the PR.

flashlight/test/memory/CachingMemoryManagerTest.cpp
```diff
@@ -138,6 +140,46 @@ TEST_F(CachingMemoryManagerTest, OOM) {
...
+}
+
+TEST_F(CachingMemoryManagerTest, Fragmentation) {
+  testFragmentation(deviceInterface_, adapter_, true); // should OOM
+}
+
+TEST_F(CachingMemoryManagerTest, RecLimit) {
+  constexpr static size_t ONE_GB = 1 << 30;
+  // Fine set the manager in order not to recycle big tensors:
+  adapter_->setOption((unsigned)fl::CachingMemoryManager::Options::RECSIZELIMIT, std::to_string(2 * ONE_GB));
```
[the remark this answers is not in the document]
WilliamTambellini: done

flashlight/test/memory/CachingMemoryManagerTest.cpp
```diff
@@ -138,6 +140,46 @@ TEST_F(CachingMemoryManagerTest, OOM) {
   }
 }
 
+void testFragmentation(std::shared_ptr<fl::MemoryManagerDeviceInterface> deviceInterface_, std::shared_ptr<fl::CachingMemoryManager> adapter_, bool expectOOM) {
```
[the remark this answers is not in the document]
WilliamTambellini: done

flashlight/test/memory/CachingMemoryManagerTest.cpp
```diff
@@ -138,6 +140,46 @@ TEST_F(CachingMemoryManagerTest, OOM) {
...
+  af_backend b;
+  af_get_active_backend(&b);
+  if (b != AF_BACKEND_CUDA)
+    GTEST_SKIP();
```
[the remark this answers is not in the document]
WilliamTambellini: done

ext/CMakeLists.txt
WilliamTambellini: done

flashlight/memory/MemoryManagerAdapter.cpp
WilliamTambellini: done

flashlight/memory/managers/CachingMemoryManager.cpp
WilliamTambellini: done

flashlight/memory/managers/CachingMemoryManager.cpp
WilliamTambellini: done

flashlight/memory/managers/CachingMemoryManager.h
```diff
@@ -50,6 +50,9 @@ class CachingMemoryManager : public MemoryManagerAdapter {
   void addMemoryManagement(int device) override;
   void removeMemoryManagement(int device) override;
 
+  enum class Options { RECSIZELIMIT, SPLITSIZELIMIT };
```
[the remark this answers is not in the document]
WilliamTambellini: done

flashlight/test/memory/CachingMemoryManagerTest.cpp
```diff
@@ -138,6 +140,46 @@ TEST_F(CachingMemoryManagerTest, OOM) {
...
+  adapter_->printInfo("After releasing a1:", 0);
+
+  af::array a2(.1f * maxNumf32);
+  adapter_->printInfo("After creating a2:", 0);
+
+  af::array a3;
+  try {
+    a3 = af::array(.5f * maxNumf32);
+  } catch (af::exception& ex) {
+    if(expectOOM)
```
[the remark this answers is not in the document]
WilliamTambellini: CRs are added by
~/repos/flashlight (CMM) $ clang-format -style=file -i flashlight/test/memory/CachingMemoryManagerTest.cpp

flashlight/test/memory/CachingMemoryManagerTest.cpp
```diff
@@ -138,6 +140,46 @@ TEST_F(CachingMemoryManagerTest, OOM) {
   }
 }
 
+void testFragmentation(std::shared_ptr<fl::MemoryManagerDeviceInterface> deviceInterface_, std::shared_ptr<fl::CachingMemoryManager> adapter_, bool expectOOM) {
+  af_backend b;
```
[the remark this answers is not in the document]
WilliamTambellini: done

flashlight/memory/managers/CachingMemoryManager.cpp
```diff
@@ -31,6 +33,12 @@ constexpr size_t kMinLargeAlloc =
     10485760; // allocations between 1 and 10 MiB may use kLargeBuffer
 constexpr size_t kRoundLarge = 2097152; // round up large allocs to 2 MiB
 
+// Non-const runtime options in order to fine tune the behavior of this manager.
+// Prevents to recycle some buffers, to be set by the user if desired:
+size_t recyclingSizeLimit = std::numeric_limits<size_t>::max();
```
WilliamTambellini, line 38: Perhaps worth to atomic.

jacobkahn (changes requested):
See suggestion here - https://github.com/facebookresearch/flashlight/pull/188#discussion_r504416826. Almost there.

flashlight/memory/managers/CachingMemoryManager.cpp
WilliamTambellini: sure, done

flashlight/memory/managers/CachingMemoryManager.cpp
```diff
@@ -126,7 +138,9 @@ void* CachingMemoryManager::alloc(
 
   CachingMemoryManager::Block* block = nullptr;
   auto it = pool.lower_bound(&searchKey);
-  if (it != pool.end()) {
+  // Recycle blocks if any found, and if small alloc or the block size is not
+  // too large:
+  if (it != pool.end() && (isSmallAlloc || (*it)->size_ < recyclingSizeLimit)) {
```
WilliamTambellini, line 143: Warning: now that recyclingSizeLimit is atomic, there could be a cost here to access/read it at each alloc.
https://stackoverflow.com/questions/2538070/atomic-operation-cost
Do you run inference/production with a high number of threads ?

jacobkahn: @WilliamTambellini — this is a good point - I actually think it's better that we don't make it atomic. I don't see a good use case for setting and getting it in multiple threads. Let's just add a comment to the doc block for the setters that says it isn't thread safe. Once that's done, we should be good to go!

WilliamTambellini: done

apply-format
```diff
@@ -0,0 +1,13 @@
+#! /bin/bash
+# Apply clang-format using the FL .clang-format on the modified/added cpp/h/hpp files of the current commit:
+clang-format --version || exit 1
+if [ ! -f .clang-format ]; then
+  echo "Cannot find .clang-fornmat in the current dir" 
+  exit 1
+fi
+for file in `git show --name-only -1 --pretty="format:" | grep -E '\.(cpp|h|hpp)$'`; do
```
jacobkahn, line 8: This is nice - would you mind sending this in a separate PR? There are a few tweaks we'd need to make to this to make sure it touches all of the potential source files :)

ext/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ include(${CMAKE_CURRENT_LIST_DIR}/common/CMakeLists.txt)
  
 # --------------------------- Tests ---------------------------
 
+# The SeqBuilder test needs some contribs modules
```
jacobkahn, line 10: nit: remove

flashlight/memory/managers/CachingMemoryManager.h
```diff
@@ -133,6 +139,14 @@ class CachingMemoryManager : public MemoryManagerAdapter {
 
   void tryMergeBlocks(Block* dst, Block* src, BlockSet& freeBlocks);
   void freeBlock(Block* block);
+
+ private:
+  // Non-const runtime options in order to fine tune the behavior of this
+  // manager. Prevents to recycle some buffers, to be set by the user if
+  // desired:
+  size_t recyclingSizeLimit;
```
jacobkahn, line 147: Our convention is to have private members' names end in underscores - e.g. `recyclingSizeLimit_`. Can you also move the initialization to inline here (e.g. `recyclingSizeLimit_{std::numeric_limits<size_t>::max()}` rather than the ctor?

jacobkahn (changes requested):
Nice — almost there! :)

apply-format
WilliamTambellini: done

ext/CMakeLists.txt
WilliamTambellini: done

flashlight/memory/managers/CachingMemoryManager.h
WilliamTambellini: done

WilliamTambellini (commented):
done

jacobkahn (approved):

jacobkahn:
@WilliamTambellini - thank again for the contribution - looks good to me - do me a favor and rebase on matter then we'll be ready to merge.

WilliamTambellini:
> @WilliamTambellini - thank again for the contribution - looks good to me - do me a favor and rebase on matter then we'll be ready to merge.

done

app/asr/CMakeLists.txt
```diff
@@ -87,9 +87,9 @@ add_executable(Train ${CMAKE_CURRENT_LIST_DIR}/Train.cpp)
 add_executable(Test ${CMAKE_CURRENT_LIST_DIR}/Test.cpp)
 add_executable(Decoder ${CMAKE_CURRENT_LIST_DIR}/Decode.cpp)
 
-target_link_libraries(Train flashlight-app-asr)
-target_link_libraries(Test flashlight-app-asr)
-target_link_libraries(Decoder flashlight-app-asr)
+target_link_libraries(Train flashlight-app-asr ${CMAKE_DL_LIBS})
```
WilliamTambellini, line 90: @jacobkahn  This fixes circleCI link errors:
[  8%] Linking CXX executable Train
/usr/bin/ld: CMakeFiles/Train.dir/flashlight/common/Plugin.cpp.o: undefined reference to symbol 'dlclose@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libdl.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status

PR #203, opened 2020-10-13
Auto-upcast precisions for reductive autograd functions

jacobkahn:
Differential Revision: D24289584


flashlight/test/nn/ModuleTest.cpp
```diff
@@ -335,6 +513,47 @@ TEST(ModuleTest, LayerNormFwd) {
...
+TEST(ModuleTest, LayerNormFwdF16) {
+  if (!af::isHalfAvailable(af::getDevice())) {
+    GTEST_SKIP() << "Half-precision not supported on this device";
+  }
+
+  double eps = 5E-2;
+  std::vector<int> feat_axes = {3};
+  auto input = Variable(af::randu(4, 4, 3, 10, af::dtype::f16), true);
+
+  auto sample_mean = mean(input, {3});
```
WilliamTambellini, line 525: Just a warning: I faced some overflows here at f16 with some real models: 
- is '4,4,3,10' the kind of shape of your input tensors with real life models ?
- have you checked that the range of the parameters values in the input tensor is also between 0 and 1 (the randu output) ?

jacobkahn: @WilliamTambellini these are good points - the new behavior here is going to be to upcast to fp32 for operators that do accumulation if they receive fp16 input. `mean` is one of them - see the new `detail::adjustInputType` in `autograd/Functions.h`. This ensures this operation occurs in fp32 which should make the behavior the same as before (doing those in fp16 was wrong/highly error prone).

The list of functions that are being changed roughly follow these lists from [NVIDIA Apex](https://github.com/NVIDIA/apex):
- https://bit.ly/33UpSWp
- https://bit.ly/30Zv2OS
- https://bit.ly/310k8Z6

WilliamTambellini: ok cool.

Issue #213, opened 2020-10-23
nvcc fatal: Unsupported gpu architecture 'compute_30'


go-dockly:
### Bug Description
nvcc fatal   : Unsupported gpu architecture 'compute_30'

How can I pass compute_61 as nvcc flag which should be supported by my card GTX 1050

#### Reproduction Steps
[Steps to reproduce. Please include, if needed for your issue:
- Any configurations or custom setup
master branch
- Commands to run
cmake .. -DCMAKE_BUILD_TYPE=Release -DFL_BACKEND=CUDA -DFL_BUILD_DISTRIBUTED=ON -DFL_BUILD_CONTRIB=OFF -DFL_BUILD_TESTS=OFF -DFL_BUILD_EXAMPLES=OFF -DFL_LIBRARIES_USE_MKL=OFF
make -j4
make install

Error
`[  1%] Building NVCC (Device) object CMakeFiles/fl-libraries-cuda.dir/lib/sequence/criterion/cuda/fl-libraries-cuda_generated_ViterbiPath.cu.o
nvcc fatal   : Unsupported gpu architecture 'compute_30'
CMake Error at fl-libraries-cuda_generated_ViterbiPath.cu.o.Release.cmake:220 (message):
  Error generating
  /home/fb/flashlight/build/CMakeFiles/fl-libraries-cuda.dir/lib/sequence/criterion/cuda/./fl-libraries-cuda_generated_ViterbiPath.cu.o`

### Platform and Hardware
Ubuntu 20.2
GeForce GTX 1050 (455.23.05)    
CUDA: 11.1
NVCC: 11.1.74
Cmake: 3.16.3

alealv:
It's an issue because of CUDA 11. Look [here](https://github.com/facebookresearch/flashlight/issues/198)

I need to make the MR with changes.

tlikhomanenko:
@alealv would be happy if you prepare the PR with all your fixes/findings on this.

jacobkahn:
This should be fixed in the aforementioned commits. Feel free to reopen if you're still having problems with CUDA compute versions.

Issue #198, opened 2020-10-11
Compile Error: "nvcc fatal   : Unsupported gpu architecture 'compute_30'"


alealv:
### Bug Description
After correct configuration
```
cmake .. -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/opt/flashlight -DINTEL_MKL_DIR=/opt/intel/mkl -DFL_BUILD_TESTS=OFF -DKENLM_LIB=/opt/kenlm/lib/libkenlm.a -DKENLM_UTIL_LIB=/opt/kenlm/lib/libkenlm_util.a -DKENLM_MODEL_HEADER=/opt/kenlm/include -DGLOG_ROOT_DIR=/opt/glog -DGLOG_INCLUDE_DIR=/opt/glog/include
```

When building I get the following error
```
/usr/local/cuda/bin/nvcc -M -D__CUDACC__ /home/aalvarez/Projects/flashlight/app/asr/third_party/warpctc/src/reduce.cu -o /home/aalvarez/Projects/flashlight/build/CMakeFiles/warpctc.dir/app/asr/third_party/warpctc/src/warpctc_generated_reduce.cu.o.NVCC-depend -ccbin /usr/bin/cc -m64 -Dwarpctc_EXPORTS -Xcompiler ,\"-fopenmp\",\"-fPIC\",\"-O3\",\"-DNDEBUG\" -std=c++11 -gencode arch=compute_30,code=sm_30 -gencode arch=compute_35,code=sm_35 -gencode arch=compute_50,code=sm_50 -gencode arch=compute_52,code=sm_52 -gencode arch=compute_60,code=sm_60 -gencode arch=compute_61,code=sm_61 -gencode arch=compute_70,code=sm_70 -gencode arch=compute_75,code=sm_75 -gencode arch=compute_70,code=compute_70 -gencode arch=compute_75,code=compute_75 -DNVCC -I/usr/local/cuda/include -I/home/aalvarez/Projects/flashlight/app/asr/third_party/warpctc/include
nvcc fatal   : Unsupported gpu architecture 'compute_30'
CMake Error at warpctc_generated_reduce.cu.o.Release.cmake:220 (message):
  Error generating
  /home/aalvarez/Projects/flashlight/build/CMakeFiles/warpctc.dir/app/asr/third_party/warpctc/src/./warpctc_generated_reduce.cu.o
```

### Platform and Hardware
This is my GPU
```
[...]

Issue #147, opened 2020-08-21
Build error: ‘cudnnSetRNNDescriptor’ was not declared in this scope


QuentGrd:
### Bug Description
I'm trying to build flashlight from sources with Cuda backend following steps provided in the documentation but an error occurs in "/flashlight/autograd/backend/cuda/CudnnUtils.cpp". The compiler says (see full logs at the end) : 
```
error: ‘cudnnSetRNNDescriptor’ was not declared in this scope; did you mean ‘cudnnSetLRNDescriptor’?
  280 |   CUDNN_CHECK_ERR(cudnnSetRNNDescriptor(
      |                   ^~~~~~~~~~~~~~~~~~~~~
```
#### Reproduction Steps
```
cmake .. -DCMAKE_BUILD_TYPE=Release -DFLASHLIGHT_BACKEND=CUDA

make clean && make -j4
# Or just the concerned file
make flashlight/autograd/backend/cuda/CudnnUtils.o 
```

### Platform and Hardware
OS: Ubuntu 20.04 LTS
kernel: 5.4.0-42-generic
ArrayFire: 3.7.2
CUDA: 11.0
cuDNN: 8.0.1 RC2
cmake: 3.16.3
gcc/g++: 9.3.0 

#### Additional Context
Full logs:

``` 
quentin@machine:~/Documents/flashlight/build$ cmake .. -DCMAKE_BUILD_TYPE=Release -DFLASHLIGHT_BACKEND=CUDA -Wno-dev
-- ArrayFire found (include: /opt/arrayfire/include, library: ArrayFire::afcuda)
-- Could NOT find cereal (missing: cereal_INCLUDE_DIRS) 
-- cereal NOT found. Will download from source
-- CUDA found (library: /usr/local/cuda-11.0/lib64/libcudart_static.a;-pthread;dl;/usr/lib/x86_64-linux-gnu/librt.so include: /usr/local/cuda-11.0/include)
-- CUDNN libname: libcudnn.so;libcudnn.dylib;cudnn64
-- Found CUDNN: (lib: /usr/lib/x86_64-linux-gnu/libcudnn.so include: /usr/include)
-- CUDNN found (library: /usr/lib/x86_64-linux-gnu/libcudnn.so include: /usr/include)
-- Will build flashlight contrib assets.
-- Gloo not found
-- Determining NCCL version from the header file: /usr/include/nccl.h
-- NCCL_MAJOR_VERSION: 2
[...]

PR #215, opened 2020-10-29
Add compiling support for CUDA v11 and cuDNN v8

alealv:
**Original Issue**: [#198] & [#213] & [#147]

### Summary

- I updated [CMakeList.txt](https://github.com/facebookresearch/flashlight/blob/master/app/asr/third_party/warpctc/CMakeLists.txt) of third party library (warp-ctc)[https://github.com/baidu-research/warp-ctc]. The latest CMakeLists.txt  handles conditional compilation for architecture **compute_30** depending on CUDA version (CUDAv11 deprecates this architecture). 
- The same is done for main [CMakeLists.txt]() file
- The [deprecation policy of CUDA has changed](https://docs.nvidia.com/deeplearning/cudnn/developer-guide/index.html#backward-compatibility). Therefore, the condition of **cudnnSetRNNDescriptor** used [here](https://github.com/facebookresearch/flashlight/blob/master/flashlight/autograd/backend/cuda/CudnnUtils.cpp#L281) needs to also include until version 8000 of cuDNN.
- Finally, with CUDA v11 [Nvidia cub](https://nvlabs.github.io/cub/) is included in CUDA toolkit, but because **flashlight** downloads and uses it's own version, it collides with _Thrust_ version. Hence,  **THRUST_IGNORE_CUB_VERSION_CHECK** must be define in if using CUDA v11

### DIFF
--- a/flashlight/fl/autograd/backend/cuda/CudnnUtils.cpp
+++ b/flashlight/fl/autograd/backend/cuda/CudnnUtils.cpp
@@ -278,7 +278,7 @@ RNNDescriptor::RNNDescriptor(
   cudnnRNNAlgo_t algo = CUDNN_RNN_ALGO_STANDARD;
   cudnnDataType_t cudnntype = cudnnMapToType(type);
 
-#if CUDNN_VERSION >= 7000
+#if CUDNN_VERSION >= 7000 && CUDNN_VERSION < 8000 
   CUDNN_CHECK_ERR(cudnnSetRNNDescriptor(
       handle,
       descriptor,


app/asr/third_party/warpctc/CMakeLists.txt
```diff
@@ -6,53 +6,87 @@ ENDIF()
...
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2")
-    # set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -O2")
-    add_definitions(-DAPPLE)
-ENDIF()
 
 include_directories(include)
 
 FIND_PACKAGE(CUDA 6.5)
-MESSAGE(STATUS "warpctc: cuda found ${CUDA_FOUND}")
+FIND_PACKAGE(Torch)
```
jacobkahn, line 13: Would it be possible to disable looking Torch up front or add `QUIET` to this at the very least? It would probably confuse the user quite a lot if the build looks for Torch.

lib/CMakeLists.txt
```diff
@@ -82,6 +82,11 @@ if (FL_LIBRARIES_USE_CUDA)
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/..> # TODO(jacobkahn): fix me
     )
 
+  # TODO For CUDA version 11 CUB doesn't have to be built 
+  if (CUDA_VERSION VERSION_GREATER "11.0" OR CUDA_VERSION VERSION_EQUAL "11.0")
```
jacobkahn, line 86: Could you potentially gate https://github.com/facebookresearch/flashlight/blob/cc7f8073b719f1d7aba289ba0e50ec9e94dca5db/lib/CMakeLists.txt#L55 with this check as well so we don't download Cub if we've found CUDA 11?

lib/CMakeLists.txt
```diff
@@ -82,6 +82,11 @@ if (FL_LIBRARIES_USE_CUDA)
...
+    add_compile_definitions(THRUST_IGNORE_CUB_VERSION_CHECK)
```
jacobkahn, line 87: Let's use `target_compile_definitions` on the `lib` target. We can make the modification `PUBLIC` so it works with libraries that link to `lib`.

jacobkahn (changes requested):
Thanks for the PR @alealv! A few quick requests

alealv:
I took the time to update CMakeLists.txt of the warpctc project and submit a [PR](https://github.com/baidu-research/warp-ctc/pull/165). It works but I think it will take time to be approved.

Because of that and because many of the project files used in *flashlight* were also outdated, I updated everything and made a PR for that.

Hence, that PR should be merge before this one.

lib/CMakeLists.txt
alealv: done

lib/CMakeLists.txt
alealv: I've tried to do that, but it didn't work, and I didn't have the time to dig into it.

alealv: Also, the latest version of [FindCudaToolkit](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#module:FindCUDAToolkit) (which should be updated in flashlight CMakeLists.txt) doesn't compile *cub* library yet, so we need to do it by hand.

tlikhomanenko:
Could you fix the build?

alealv:
> Could you fix the build?

Fixed! [#255]

jacobkahn:
@alealv — thanks for this.

I'm actually about to commit some changes that update our CMake minimum version to 3.10 which has much better CUDA support (as some of us have been discussing in another thread). Once that's in, it would be great if you could rebase on top of that, then I'll review this.

alealv:
Given that most of the changes were added in #266, I updated this PR to contain only the following change that is missing and stoping the master branch to compile if CUDAv11 is present:

- The [deprecation policy of CUDA has changed](https://docs.nvidia.com/deeplearning/cudnn/developer-guide/index.html#backward-compatibility). Therefore, the condition of **cudnnSetRNNDescriptor** used [here](https://github.com/facebookresearch/flashlight/blob/master/flashlight/autograd/backend/cuda/CudnnUtils.cpp#L281) needs to also include until version 8000 of cuDNN.

jacobkahn:
Thanks for those updates @alealv — especially thanks for continuing to rebase on top of recent changes (things are changing quickly) :) - taking a look

Issue #205, opened 2020-10-15
Error compiling Python Bindings


alealv:
### Bug Description
When compiling Python Bindings, more specific `_dictionary.cpp` I get the following error:

```
/usr/lib/ccache/c++ -fPIC  -fopenmp -O3 -DNDEBUG -shared  -o _lib_text_dictionary.cpython-38-x86_64-linux-gnu.so CMakeFiles/_lib_text_dictionary.dir/bindings/python/flashlight/lib/text/_dictionary.cpp.o  -Wl,-rpath,/opt/intel/mkl/lib/intel64 -flto libfl-libraries.a /opt/intel/mkl/lib/intel64/libmkl_gf_lp64.so /opt/intel/mkl/lib/intel64/libmkl_gnu_thread.so /opt/intel/mkl/lib/intel64/libmkl_core.so -liomp5 -lm -lfftw3 -lfftw3f -lfftw3l -lgomp -lpthread /opt/kenlm/lib/libkenlm.a /opt/kenlm/lib/libkenlm_util.a -llzma -lbz2 -lz libfl-libraries-cuda.a -pthread /usr/local/cuda/lib64/libcudart_static.a -ldl -lrt /usr/local/cuda/lib64/libcudart_static.a -ldl -lrt 
/usr/bin/ld: libfl-libraries.a(Dictionary.cpp.o): relocation R_X86_64_PC32 against symbol `_ZNKSt5ctypeIcE8do_widenEc' can not be used when making a shared object; recompile with -fPIC
``` 
But the thing is I'm correctly passing `-fPIC` at the beginning of the line

This is my Cmake configuration
```
[...]

PR #225, opened 2020-10-30
Update warpctc library

alealv:
**Original PR**:  [!215]

### Summary
Brings the latest updates of *warpctc* and uses warpctc CMakeLists.txt submit in [this PR](https://github.com/baidu-research/warp-ctc/pull/165). It works but I think it will take time to be approved.


alealv:
The only problem I have is that I got this warning when configuring CMake
```
CMake Error in flashlight/app/asr/third_party/warpctc/CMakeLists.txt:
  Target "warpctc" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "/home/aalvarez/Projects/flashlight/flashlight/app/asr/third_party/warpctc/include"

  which is prefixed in the source directory.
```

alealv:
Now it's working without issues

tlikhomanenko:
Could you fix the build?

jacobkahn (changes requested):
CI is still failing because `warpctc` isn't in the export set for installing the asr lib target. Can we check if you're commenting out a line that might affect `INSTALLABLE_TARGETS`?

flashlight/app/asr/third_party/warpctc/src/ctc_entrypoint.cpp
```diff
@@ -27,8 +27,7 @@ const char* ctcGetStatusString(ctcStatus_t status) {
         return "invalid value";
     case CTC_STATUS_EXECUTION_FAILED:
         return "execution failed";
-    case CTC_STATUS_LABEL_LENGTH_TOO_LARGE:
```
vineelpratap, line 30: please keep this check

flashlight/app/asr/third_party/warpctc/include/detail/gpu_ctc_kernels.h
```diff
@@ -365,7 +365,7 @@ void compute_betas_and_grad_kernel (const ProbT* probs, const int *label_sizes,
             // Update input buffer for next iteration
             if ((tid == 0) && (end == S))
                 temp_buffer.beta[(S-1)] = temp_buffer.beta[(S-1)] +
-                                          probs[prob_offset + start_prob_col + blank_label];
+                                          log(probs[prob_offset + start_prob_col + blank_label]);
```
vineelpratap, line 368: Some of the changes here were made intentionally to make warp-ctc more numerically stable. Could you keep those changes.

vineelpratap: See https://github.com/facebookresearch/wav2letter/commit/35c449f31cd97e9ae408544d51893c879fd56063#diff-64488da267bc453d8e75d64ee329b2508486ee8b1345ff59af93d38db644b760 

https://github.com/facebookresearch/wav2letter/commit/7258c1697bbcc30ca0c3e06718ee62203a2c33d4#diff-64488da267bc453d8e75d64ee329b2508486ee8b1345ff59af93d38db644b760

vineelpratap (changes requested):

alealv:
> CI is still failing because `warpctc` isn't in the export set for installing the asr lib target. Can we check if you're commenting out a line that might affect `INSTALLABLE_TARGETS`?

`warpctc` is being set in [flashlight/app/CMakeLists.txt](https://github.com/alealv/flashlight/blob/update_warp_ctc/flashlight/app/CMakeLists.txt). I think that the issue is because this file is not the one with the `add_subdirectory()` sentence, instead is being called in [flashlight/app/asr/criterion/CMakeLists.txt](https://github.com/alealv/flashlight/blob/update_warp_ctc/flashlight/app/asr/criterion/CMakeLists.txt). Although, I don't know how should I properly fixed it.

flashlight/app/asr/third_party/warpctc/src/ctc_entrypoint.cpp
alealv: done

flashlight/app/asr/third_party/warpctc/include/detail/gpu_ctc_kernels.h
alealv: done

flashlight/app/asr/criterion/CMakeLists.txt
```diff
@@ -13,10 +13,10 @@ if (FLASHLIGHT_USE_CUDA)
   # HORRENDOUS HACK (remove me when we remove warpctc):
   # see https://cmake.org/Bug/view.php?id=14444
   add_subdirectory(${WARPCTC_DIR} ${PROJECT_BINARY_DIR}/third_party/warpctc)
-  cuda_add_library(warpctc SHARED ${warpctc_SOURCES})
-  target_link_libraries(warpctc ${warpctc_LIBS})
-  set(WARPCTC_INCLUDE_DIR ${WARPCTC_DIR}/include)
-  target_include_directories(warpctc PRIVATE ${WARPCTC_INCLUDE_DIR})
+  #cuda_add_library(warpctc SHARED ${warpctc_SOURCES})
```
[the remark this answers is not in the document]
alealv: done

alealv:
I found the problem. It's the version of cmake. I will update it in the docker file in the next commit.

Also, the dockerfiles are quite old and they use Ubuntu 16.04. I will make a PR updating to the latest with CUDA 11.1

alealv:
I just updated everything, but I realized that the docker image isn't built in the CI. Should I update this?

tlikhomanenko:
Hi @alealv, thanks for your questions, please find my answers below.

> I found the problem. It's **cmake** version. I will update it in the docker file in the next commit.

Thanks!

> BTW, I have many questions regarding your CI and Dockerfile:
> 
> * Why are you compiling SoundFile? Ubuntu has [it already](https://packages.ubuntu.com/xenial/libsndfile1-dev)

Good point, maybe @jacobkahn knows why we did this. We can try to use standard ubuntu stuff. Feel free to send PR if you have time to test it =).

> * Why do you have everything in a single `RUN` instruction? This is doesn't keep intermediate caches if something is wrong and then you need to start from the beginning again.

When I test it locally I do by hand insertions of RUN and then debug. As final version we do all in 1 run because if you don't cleanup cache, tmp, etc. this will be stored in the intermediate layer and thus you will have larger size of final image. Because docker RUN works as an additional layer so it doesn't change the size of previous layer with RUN command. Our docker images are huge already, so this is just an attempt to have a bit less size.

> * Why are you compiling all the tools with the default version of GCC for the OS and then installing GCCv5 for compiling flashlight? This can lead to failure.

Good point, but mainly we have this on CI to be sure that for gcc 5 it is still working )people reported a lot of building issues some time for gcc v5), but docker images we have with latest version which is in Ubuntu. Let me think how we can fix this.

> * Why are you installing CUDA on the virtual machine? You only need to install the **nvidia drivers**, **docker** and **nvidia-docker2**

can you tell more how you propose to install nvidia driver (at least with the way we do it installs cuda driver too anyway)?
 
> * Why aren't you compiling flashlight inside the Dockerfile, but doing it as an external command?

for both CPU and CUDA CI we do compilation inside the docker files. We could just run Dockerfile build (where only flashlight builds) and do fix on test running to avoid duplication of flashlight build commands. Thanks for pointing. I need to check a bit how to do so for CI.

> Finally, why don't you publish a Docker with flashlight in Dockerhub? I guess a lot of people will use it.

in .docker folder you can see that we have CPU and GPU base dockerfiles where we build all dependencies, and then dockerfiles which builds flashlight. We have github actions which build docker image of flashlight on top of base images and push it to dockerhub. So you can download both base images and image with flashlight build on top.

tlikhomanenko:
> I just updated everything, but I realized that the docker image isn't built in the CI. Should I update this?

I need to upload updated version of base docker image, on top of which we build the flashlight. Let me test this locally and update base image (should be fast, as you tested that it is working for you). As review of PR will be done and base image is update we can rerun tests in CI and then merge. So let us review your PR now.

alealv:
> Good point, maybe @jacobkahn knows why we did this. We can try to use standard ubuntu stuff. Feel free to send PR if you have time to test it =).

I've been using **flashlight** with **fairseq** for training Wav2Vec 2.0 and I didn't have any issue with audio data loading.

> When I test it locally I do by hand insertions of RUN and then debug. As final version we do all in 1 run because if you don't cleanup cache, tmp, etc. this will be stored in the intermediate layer and thus you will have larger size of final image. Because docker RUN works as an additional layer so it doesn't change the size of previous layer with RUN command. Our docker images are huge already, so this is just an attempt to have a bit less size.

I think the best would be to use a Multistage build, but also you could use the experimental option `--squash` (check [this post](https://medium.com/@gdiener/how-to-build-a-smaller-docker-image-76779e18d48a)).
Also for a final image with flashlight installed the size can be reduce a lot, a mean a lot, if you use as base (after building) `cuda:XX.X-cudnnX-runtime-ubuntuXX.XX` instead ([for example](https://hub.docker.com/layers/nvidia/cuda/11.1-cudnn8-runtime-ubuntu18.04/images/sha256-1400f0c464e7dc77b1c420976410900fc4e04cc98e5a19c9824ce057282bfc53?context=explore))

> Good point, but mainly we have this on CI to be sure that for gcc 5 it is still working (people reported a lot of building issues some time for gcc v5), but docker images we have with latest version which is in Ubuntu. Let me think how we can fix this.

Personally, I think gcc v5 is quite old but I see that Ubuntu16.04 has at most GCC v5.3.1

> can you tell more how you propose to install nvidia driver (at least with the way we do it installs cuda driver too anyway)?

You can simply run `sudo apt-get -y install cuda-drivers`
Check [here](https://docs.nvidia.com/datacenter/tesla/tesla-installation-notes/index.html#ubuntu-lts) and/or [here](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#driver-installation)

tlikhomanenko:
> > Good point, maybe @jacobkahn knows why we did this. We can try to use standard ubuntu stuff. Feel free to send PR if you have time to test it =).
> 
> I've been using **flashlight** with **fairseq** for training Wav2Vec 2.0 and I didn't have any issue with audio data loading.
> 
> > When I test it locally I do by hand insertions of RUN and then debug. As final version we do all in 1 run because if you don't cleanup cache, tmp, etc. this will be stored in the intermediate layer and thus you will have larger size of final image. Because docker RUN works as an additional layer so it doesn't change the size of previous layer with RUN command. Our docker images are huge already, so this is just an attempt to have a bit less size.
> 
> I think the best would be to use a Multistage build, but also you could use the experimental option `--squash` (check [this post](https://medium.com/@gdiener/how-to-build-a-smaller-docker-image-76779e18d48a)).
> Also for a final image with flashlight installed the size can be reduce a lot, a mean a lot, if you use as base (after building) `cuda:XX.X-cudnnX-runtime-ubuntuXX.XX` instead ([for example](https://hub.docker.com/layers/nvidia/cuda/11.1-cudnn8-runtime-ubuntu18.04/images/sha256-1400f0c464e7dc77b1c420976410900fc4e04cc98e5a19c9824ce057282bfc53?context=explore))
> 
> > Good point, but mainly we have this on CI to be sure that for gcc 5 it is still working (people reported a lot of building issues some time for gcc v5), but docker images we have with latest version which is in Ubuntu. Let me think how we can fix this.
> 
> Personally, I think gcc v5 is quite old but I see that Ubuntu16.04 has at most GCC v5.3.1
> 
> > can you tell more how you propose to install nvidia driver (at least with the way we do it installs cuda driver too anyway)?
> 
> You can simply run `sudo apt-get -y install cuda-drivers`
> Check [here](https://docs.nvidia.com/datacenter/tesla/tesla-installation-notes/index.html#ubuntu-lts) and/or [here](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#driver-installation)

Thanks for your comments and suggestions, added in my todo on fixing this. But will work on this in several weeks only probably. About gcc - yep, mostly we are support the version which is default in ubuntu 16 for now, as people are still widely using it.

WilliamTambellini:
> Personally, I think gcc v5 is quite old but I see that Ubuntu16.04 has at most GCC v5.3.1

If I may comment: the question is not really into the version of gcc, but more into the c++ standard: at the moment flashlight only  allows c++11 in the main cmake: 
```
set_target_properties(
    flashlight
    PROPERTIES
    LINKER_LANGUAGE CXX
    CXX_STANDARD 11
    )
```
at least for the fl core cpp side (not speaking about third parties). A step fwd would be to adopt cpp14.
gcc 5 seems to be c++14 capable:
https://gcc.gnu.org/projects/cxx-status.html
@tlikhomanenko  any reason to restrict to cpp 11 ?

.docker/Dockerfile-CPU-Base
```diff
@@ -3,7 +3,7 @@
 # ------------------------------------------------------------------
 # Ubuntu           18.04
 # OpenMPI          latest       (apt)
-# cmake            3.10         (git)
+# cmake            3.18         (git)
```
avidov, line 6: 3.18 is too new. Ubuntu 18.04 supports 3.10
https://launchpad.net/ubuntu/bionic/+source/cmake
Ubuntu 20.04 supports 3.16
https://launchpad.net/ubuntu/focal/+source/cmake
I think that we don't want to break the builds for users on these platforms.

alealv: [Snapcraft](https://snapcraft.io/cmake) which is installed by default in Ubuntu has CMake 3.18.4 as default and is super easy to install:

```sh
snap install --classic cmake
```

And even so, you could download the [compiled version for Linux ](https://github.com/Kitware/CMake/releases/download/v3.18.4/cmake-3.18.4-Linux-x86_64.tar.gz) from their repo and it works out of the box. And installing from source is also super easy.

We could try to run it with CMake 3.16, but I think the minimum version is 3.15 although I haven't tested.

jacobkahn:
Hey all -- I'm sorry I'm late to this discussion. Thanks for all of this conversation so far. I'll answer some of the questions and will post another response later when I'm back at my machine. Please do add any follow-up questions from these. I hope this information is generally useful to everyone, cc @WilliamTambellini @tlikhomanenko @avidov @padentomasello.

### This PR:

@aleav — thanks again for your work on this and for all of your input. I will get to reviewing this change soon (see my note below as well) I'm going to answer a bunch of questions as best I can here:

> Why are you compiling SoundFile? Ubuntu has it already

As it stands, for flashlight, `libsndfile` file needs to be built with the "external libraries" functionality and the last time I wnt through our Docker config, the apt packages don't have those things linked. Installing `libogg`, `libflac`, libvorbis` before building `libsndfile` from source ensures that it will be built with those external libraries.

If things in `apt` are updated, we won't need to build from source anymore. I haven't checked this in a while but will check again.

> Why do you have everything in a single RUN instruction? This is doesn't keep intermediate caches if something is wrong and then you need to start from the beginning again.

Absolutely agree -- this should definitely be changed. Things are really hard to reason about with a single `RUN`. I'm in the process of splitting up our CI components to run tests from different flashlight components, and I've done this as part of that change, so expect to see this soon.

I think the original reason we did this was because CircleCI didn't have the ability to run `docker run` with `--runtime=nvidia`, which ensures that the guest OS has access to the NVIDIA runtime. I'll do a check to see if there's updated wisdom about this, but either way, we can probably split up our `RUN`s. Thanks for pointing this out either way.

> Why are you compiling all the tools with the default version of GCC for the OS and then installing GCCv5 for compiling flashlight? This can lead to failure.

Yep. It's bad. I also have a draft fix for this which makes the CI turnaround significantly better. I hope to get that out early this week.

> Why are you installing CUDA on the virtual machine? You only need to install the nvidia drivers, docker and nvidia-docker2

Best I can remember, `nvidia-docker 2.0` wasn't around when we originally worked on this, but I'll look into this again. CUDA was required at the time to do what we needed. Thanks for the pointer.

> Why aren't you compiling flashlight inside the Dockerfile, but doing it as an external command?

I don't quite understand this question -- we want to do do a build from source -- the Dockerfile we use is a "base" dockerfile which has all of the dependencies installed except for flashlight. This ensures dependencies don't break the CI build randomly.

### Re CUDA and flashlight:

I'm actually in the process of refactoring a lot of our build components that deal with CUDA things to enable separable compilation and device linking, so some of the changes in this PR will probably need to be rebased on that. The goal is to link warpctc and the flashlight library CUDA kernels into a single shared object so that we aren't exporting a million CMake targets.

As part of this, I'll further centralize the place where we configure `nvcc`. The warpctc CMake stuff we have is a disaster because we have multiple sources of truth adding flags to the compiler, and old CMake versions don't support CUDA library targets, so warpctc's nvcc flags are getting passed to global context. Part of the work I'm doing is consolidating the places where we configure nvcc and CUDA compilation so random options in warpctc don't implicate how we build CUDA kernels in the rest of flashlight and cause inconsistencies.

### Why are we still C++11 compatible? Why this gcc >= 5 requirement?

We've had discussions on this internally, and I'm thinking of relaxing this soon. The reasons for wanting to stick to old compiler versions are pretty straightforward -- the biggest is that we want to make sure that people can build flashlight on Ubuntu 14.04 with the default compiler, but given that 18.04 is quite standard now (not to mention 16.04) following default compilers 

### Why support older versions of CMake?

This has similar logic to the above - we want to follow aptitude ppas for some minimum Ubuntu version, which was 14.04. We'll see about relaxing this. I should revisit the the closure of minimum CMake versions for our dependencies and see if we can't bump things -- ArrayFire still supports 3.5.1 as a minimum.

## An important note:

`vcpkg` support for flashlight (and everything in our dependency tree) is on the final stretch which will hopefully make setting up an environment significantly easier. Using `vcpkg` in CI for flashlight's dependencies will further simplify our CI if we choose to go that route.

alealv:
Thanks, @jacobkahn, and @tlikhomanenko for your clear explanations.

**Regarding my question**:
> I don't quite understand this question -- we want to do do a build from source -- the Dockerfile we use is a "base" dockerfile which has all of the dependencies installed except for flashlight. This ensures dependencies don't break the CI build randomly.

I meant that you are starting a docker container and running [these lines](https://github.com/facebookresearch/flashlight/blob/master/.circleci/config.yml#L63) inside, instead of executing docker build of the [Dockerfile-CUDA](https://github.com/facebookresearch/flashlight/blob/master/.docker/Dockerfile-CUDA) and [Dockerfile-CPU](https://github.com/facebookresearch/flashlight/blob/master/.docker/Dockerfile-CPU).

**Regarding this**:
> If things in apt are updated, we won't need to build from source anymore. I haven't checked this in a while but will check again.

If you check [all dependencies](https://packages.ubuntu.com/bionic/libsndfile1-dev
) are satisfied.

**Regarding docker**: I already have a [branch in my fork](https://github.com/alealv/flashlight/tree/update-Docker-images/.docker) where I updated the Docker images with two stages, which reduces the sizes a lot (I don't have the numbers now but it gets reduce almost to half). I was going to do a PR later, but you can use it as inspiration if you already have something in mind.
I also add a `.dockerignore` that reduces context loading time and I'm working on getting more stages to make a faster build activating [Docker Built Kit](https://docs.docker.com/develop/develop-images/build_enhancements/).

**Regarding CMake**: First, since version 3.10 the CUDA language has been integrated into CMake and it's supported directly, so it would be better using regular CMake commands. But secondly, I would suggest moving at least to version 3.17, because now [FindCUDA](https://cmake.org/cmake/help/latest/module/FindCUDA.html) is deprecated in favor of [FindCUDAToolkit](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html). I guess in the near future they will add support for [cub](https://github.com/NVIDIA/cub) (a requirement for flashlight), which is integrated with [CUDA toolkit since version 11](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#major-component-descriptions).

Pushing a little further, I think it would be also great to [fix this](https://github.com/facebookresearch/flashlight/issues/205), and I would like for any of you to review this [PR](https://github.com/pytorch/fairseq/pull/2876) I did into fairseq. Sorry if I'm pushing too much, but I've been trying to fine-tune Wav2vec 2.0 in the last weeks and it's been complicated with the actual state of things.

jacobkahn:
@alealv —

> CMake version

I've decided to take this one step at a time and upgrade to 3.10 first -- https://github.com/facebookresearch/flashlight/pull/266. That PR has more context on the change. That should be committed today or tomorrow and will probably allow you to eliminate a lot of code from this PR.

> instead of executing docker build

My understanding is that executing `Docker build` on those would also build the dependencies from source which isn't what we want - it takes too long. That said, I'll look into restructuring this based on your suggestions. Thanks.

> If you check [all dependencies](https://packages.ubuntu.com/bionic/libsndfile1-dev) are satisfied.

I don't see `ogg` as a dependency, so it's unclear to me if this library is built with those external features. Either way, I'll give it a try soon.

> I was going to do a PR later, but you can use it as inspiration if you already have something in mind.

These sound like great changes — thanks so much for those. Even if they're incomplete, sending incremental PRs so we can test along the way would be very helpful. Please do share the branch.

> Pushing a little further, I think it would be also great to [fix this](https://github.com/facebookresearch/flashlight/issues/205), and I would like for any of you to review this [PR](https://github.com/pytorch/fairseq/pull/2876)

@tlikhomanenko would you be able to review https://github.com/facebookresearch/flashlight/issues/205? I can also take a look soon.

alealv:
> I've decided to take this one step at a time and upgrade to 3.10 first -- #266. That PR has more context on the change. That should be committed today or tomorrow and will probably allow you to eliminate a lot of code from this PR.

Cool, I will look into that

> My understanding is that executing `Docker build` on those would also build the dependencies from source which isn't what we want - it takes too long. That said, I'll look into restructuring this based on your suggestions. Thanks.

No. You already have [these lines](https://github.com/facebookresearch/flashlight/blob/master/.circleci/config.yml#L58) to log into your docker registry. If the base image is up there it will pull it, otherwise, it will fail, unless it has it cached on the machine (but I think this is not the case).

> I don't see `ogg` as a dependency, so it's unclear to me if this library is built with those external features. Either way, I'll give it a try soon.

Because it's a dependency from [liborbis-dev](https://packages.ubuntu.com/bionic/libvorbis-dev), so it will be installed.

> These sound like great changes — thanks so much for those. Even if they're incomplete, sending incremental PRs so we can test along the way would be very helpful. Please do share the branch.

Cool

tlikhomanenko:
> @alealv —
> 
> > CMake version
> 
> I've decided to take this one step at a time and upgrade to 3.10 first -- #266. That PR has more context on the change. That should be committed today or tomorrow and will probably allow you to eliminate a lot of code from this PR.
> 
> > instead of executing docker build
> 
> My understanding is that executing `Docker build` on those would also build the dependencies from source which isn't what we want - it takes too long. That said, I'll look into restructuring this based on your suggestions. Thanks.

No we really can just run docker build on CPU or CUDA and with building it will run loading base image and then building. This will definitely remove duplication. @jacobkahn I can help on this if you want.

> 
> > If you check [all dependencies](https://packages.ubuntu.com/bionic/libsndfile1-dev) are satisfied.
> 
> I don't see `ogg` as a dependency, so it's unclear to me if this library is built with those external features. Either way, I'll give it a try soon.
> 
> > I was going to do a PR later, but you can use it as inspiration if you already have something in mind.
> 
> These sound like great changes — thanks so much for those. Even if they're incomplete, sending incremental PRs so we can test along the way would be very helpful. Please do share the branch.
> 
> > Pushing a little further, I think it would be also great to [fix this](https://github.com/facebookresearch/flashlight/issues/205), and I would like for any of you to review this [PR](https://github.com/pytorch/fairseq/pull/2876)
> 
> @tlikhomanenko would you be able to review #205? I can also take a look soon.

Yep, we had discussion before with @alealv, but didn't have time to test more intensively. I think we should finish first with vcpkg and all move to 3.10 cmake, so that the main cmake is set, and then I can work on python bindings separately, does this sound good @jacobkahn @alealv (for now let's just be sure that some of the installation way is working for python bindings)?

alealv:
I close this issue because it's already being updated with #266

tlikhomanenko:
Python bindings improvement (with python setup.py install) https://github.com/facebookresearch/flashlight/pull/404

PR #240, opened 2020-11-04
Use imported targets for audio encoding libs

jacobkahn:
Differential Revision: D24720157


jacobkahn:
[Doesn't pass CI -- this is a separate commit because the case of `FindSNDFILE.cmake` had to change -- two commits are required. See https://github.com/facebookresearch/flashlight/pull/238 CI for signal].

PR #244, opened 2020-11-06
Conformer step 2: its implementation

tlikhomanenko:
### Summary
Add conformer implementation (similar to the original paper. Main difference is i) another relative positional embedding ii) batchnorm -> layernorm

### Test Plan (required)
tested with 120M param model on convergence, final WER is 1% far away from the reported in the paper (I didn't use RNN-T; probably different word-piece set).

### DIFF
--- a/flashlight/fl/contrib/modules/Conformer.cpp
+++ b/flashlight/fl/contrib/modules/Conformer.cpp
@@ -1,0 +1,174 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+#include "flashlight/fl/contrib/modules/Conformer.h"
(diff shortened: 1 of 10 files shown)


flashlight/fl/contrib/modules/Conformer.cpp
```diff
@@ -0,0 +1,176 @@
...
+Variable Conformer::conv(const Variable& input) {
+  float pDropout = train_ ? pDropout_ : 0.0;
+  // input C x T x B x 1
+  auto result = reorder(input, 1, 3, 0, 2);
+  // T x 1 x C x B
+  // apply first pointwise conv
+  result =
+      gatedlinearunit((*conv1_)(((*normConv1_)(result)).as(input.type())), 2);
+  // apply depthwise separable convolutions
+  result = (*convDepthWiseStep2_)((*convDepthWiseStep1_)(result));
```
syhw, line 123: I'd try `result = (*convDepthWiseStep1_)(result);`

PR #247, opened 2020-11-08
Fix RnnLm example with missing LogSoftmax fwd

### LINKED ISSUE #139: missing logsoftmax operation in example Rnnlm
in example/Rnnlm, we define a class Rnnlm with forward function:
  std::tuple<Variable, Variable, Variable>
  forward(const Variable& input, const Variable& h, const Variable& c) {
    auto output = embed(input);
    Variable ho, co;
    std::tie(output, ho, co) = rnn(output, h, c);

    // Truncate BPTT
    ho.setCalcGrad(false);
    co.setCalcGrad(false);

    output = linear(output);
    return std::make_tuple(output, ho, co);
  }
and use CategoricalCrossEntropy to calc loss, but CategoricalCrossEntropy need logsoftmax as input, so: Maybe this is a bug! 

RnnLm::forward  function ignores operation: logsoftmax, which should use after linear and before return.

WilliamTambellini:
Add a missing logsoftmax fwd in the RnnLm example model
to comply with the CatCrossEntropy loss.
fix #139
https://github.com/facebookresearch/flashlight/issues/139

**Original Issue**:
https://github.com/facebookresearch/flashlight/issues/139

closes #139

### Summary
Add a missing logsoftmax fwd in the RnnLm example model
to comply with the CatCrossEntropy loss.

### Test Plan (required)
Run the example and see normal/positive/acceptable positive decreasing loss.

### DIFF
--- a/flashlight/fl/examples/RnnLm.cpp
+++ b/flashlight/fl/examples/RnnLm.cpp
@@ -106,6 +106,8 @@ class RnnLm : public Container {
     co.setCalcGrad(false);
 
     output = linear(output);
+    LogSoftmax lsm(0); // max on the main dimension
+    output = lsm(output);
     return std::make_tuple(output, ho, co);
   }
 


tlikhomanenko:
Thanks for the PR and fix!

PR #266, opened 2020-11-17
Upgrade to CMake 3.10 minimum, shared lib and CUDA support

jacobkahn:
Summary:
*Looking for comments here.* We don't have to go with this, but this is a draft of what this looks like. Helps us eliminate a lot of technical debt for now. We might be able to one day relax this restriction down the line again if we need to.

Update the minimum version of CMake required for flashlight to 3.10. 3.10 deprecates `FindCUDA.cmake` and adds first-class support for CUDA code in CMake as a project language.

Among other things, this diff:
- **Adds support for CUDA 11** and running flashlight code on Ampere with the CMake build
- **Adds support for building flashlight libraries, flashlight core, and flashlight apps as shared objects/libraries**. This makes shipping these libraries to other places easy.
- **Removes the `fl-libraries-cuda` and `warpctc` targets from flashlight**. These used to be separate CUDA libraries that were linked with and exported as additional targets
  - They made understanding what flashlight exported more difficult to understand
  - They had the potential to conflict with other bits, i.e. another build of warpctc on the user's machine
  - CUDA kernels in flashlight libraries are now part of the `fl-libraries` target which makes it transparently easy to add new CUDA sources to flashlight, cc chaitan3
- CMake 3.10 automatically enables CUDA Separable Compilation, device linking, and PIC.
- Removes the backport of `cuda_add_library` that was necessary for older versions of CMake
- Builds warpctc as part of `fl-app-asr`
- Removes unused cruft in warpctc related to the warpctc CPU backend (which we will never use or build)
- Removes direct linking of `OpenMP` and `Threads` `IMPORTED` targets because they incorrectly propagate flags to nvcc which aren't interpreted properly. This is probably fixable in a better way.
- Gates the installation of CUB for CUDA versions <= 11 (this is necessary - including CUB with CUDA >= 11 breaks things)

Differential Revision: D25013556


alealv:
Hi, many thanks for the update!  This has simplified a lot the building procedure and code.

Although, it's incomplete because it's not building for CUAD v11. You forgot to add the changes of this [commit](https://github.com/alealv/flashlight/commit/e6d5944f38cdf8589c8577e56f39ba9454e867f1). I updated #215 to contain this.

PR #285, opened 2020-11-30
Fix non MKL build linking with OpenMP

### LINKED ISSUE #283: Linking failure on openMP symbols
### Bug Description
Linking failure when FL_LIBRARIES_USE_MKL=OFF:
```
[ 63%] Linking CXX executable MfccTest
cmake-3.13.3/bin/cmake -E cmake_link_script CMakeFiles/MfccTest.dir/link.txt --verbose=1
/usr/bin/c++     CMakeFiles/MfccTest.dir/audio/feature/MfccTest.cpp.o CMakeFiles/MfccTest.dir/cmake_device_link.o  -o MfccTest  -L/usr/local/cuda/targets/x86_64-linux/lib/stubs  -L/usr/local/cuda/targets/x86_64-linux/lib ../../../libfl-libraries.a ../../../googletest/src/gtest/./lib/libgtest.a ../../../googletest/src/gtest/./lib/libgtest_main.a ../../../googletest/src/gtest/./lib/libgmock.a ../../../googletest/src/gtest/./lib/libgmock_main.a /usr/lib/libopenblas.so /usr/lib/x86_64-linux-gnu/libfftw3.so /home/wtambellini/repos/kenlm/build/lib/libkenlm.a /home/wtambellini/repos/kenlm/build/lib/libkenlm_util.a /usr/lib/x86_64-linux-gnu/liblzma.so /usr/lib/x86_64-linux-gnu/libbz2.so /usr/lib/x86_64-linux-gnu/libz.so /usr/local/cuda/lib64/libcudart_static.a -pthread -ldl /usr/lib/x86_64-linux-gnu/librt.so -lcudadevrt -lcudart_static -lrt -lpthread -ldl 
../../../libfl-libraries.a(PowerSpectrum.cpp.o): In function `fl::lib::audio::PowerSpectrum::batchApply(std::vector<float, std::allocator<float> > const&, int)':
PowerSpectrum.cpp:(.text+0xb18): undefined reference to `GOMP_parallel'
../../../libfl-libraries.a(PowerSpectrum.cpp.o): In function `fl::lib::audio::PowerSpectrum::batchApply(std::vector<float, std::allocator<float> > const&, int) [clone ._omp_fn.0]':
PowerSpectrum.cpp:(.text+0xf62): undefined reference to `omp_get_num_threads'
PowerSpectrum.cpp:(.text+0xf6a): undefined reference to `omp_get_thread_num'
collect2: error: ld returned 1 exit status
```

#### Reproduction Steps
cmake -DFL_LIBRARIES_USE_MKL=OFF
make

### Platform and Hardware
cmake 3.10, ubunut 16, gcc5.4, 

#### Additional Context
PowerSpectrum really seems to need an openmp impl: 
as today, the master branch does not explicitly link with OpenMP::OpenMP_CXX but iomp only if MKL ON: 
```
-- Checking for [mkl_gf_lp64 - mkl_gnu_thread - mkl_core - iomp5 - pthread - m]
--   Library mkl_gf_lp64: /opt/intel/mkl/lib/intel64/libmkl_gf_lp64.so
--   Library mkl_gnu_thread: /opt/intel/mkl/lib/intel64/libmkl_gnu_thread.so
--   Library mkl_core: /opt/intel/mkl/lib/intel64/libmkl_core.so
--   Library iomp5: /usr/lib/x86_64-linux-gnu/libiomp5.so
--   Library pthread: /usr/lib/x86_64-linux-gnu/libpthread.so
--   Library m: /usr/lib/x86_64-linux-gnu/libm.so
```
gomp seems to do the job but to be used only if MKL off.
Refs:
https://github.com/pytorch/pytorch/issues/12535

WilliamTambellini:
Fix non MKL build linking with OpenMP

**Original Issue**: 
closes #283
https://github.com/facebookresearch/flashlight/issues/283

### Summary
Linking failure when MKL OFF

### Test Plan (required)
cmake -DFL_LIBRARIES_USE_MKL=OFF
make
make test

### DIFF
--- a/flashlight/lib/audio/feature/CMakeLists.txt
+++ b/flashlight/lib/audio/feature/CMakeLists.txt
@@ -48,10 +48,17 @@ target_sources(
   ${CMAKE_CURRENT_LIST_DIR}/Windowing.cpp
   )
 
+if (NOT FL_LIBRARIES_USE_MKL)
+  # openmp required by PowerSpectrum: provided by iomp if MKL ON
+  # Not using the target OpenMP::OpenMP_CXX: it does add fopenmp to nvcc which does not understand it
+  set(OMPLIB ${OpenMP_CXX_LIBRARIES})
+endif()
+
 target_link_libraries(
   fl-libraries
   PRIVATE
   ${CBLAS_LIBRARIES}
+  ${OMPLIB}
   FFTW3::fftw3
   )
 


cmake/TestUtils.cmake
```diff
@@ -18,6 +18,11 @@ function(build_test SRCFILE LINK_LIBRARIES PREPROC_DEFS)
   if (TARGET gtest)
     add_dependencies(${target} gtest) # make sure gtest is built first
   endif()
+  # openmp required by some tests using for instance PowerSpectrum: provided by iomp if MKL ON
+  if (NOT FL_LIBRARIES_USE_MKL)
+    list(APPEND LINK_LIBRARIES OpenMP::OpenMP_CXX)
```
jacobkahn, line 23: Using a global property is definitely not what we want here.

If anything, we'd want to use `target_link_libraries`, but this shouldn't be necessary here should we should be able to include these linker flags from a link call to the target that requires OpenMP in the first place.

flashlight/app/asr/CMakeLists.txt
```diff
@@ -62,9 +62,13 @@ add_executable(fl_asr_train ${CMAKE_CURRENT_LIST_DIR}/Train.cpp)
 add_executable(fl_asr_test ${CMAKE_CURRENT_LIST_DIR}/Test.cpp)
 add_executable(fl_asr_decode ${CMAKE_CURRENT_LIST_DIR}/Decode.cpp)
 
-target_link_libraries(fl_asr_train flashlight-app-asr ${CMAKE_DL_LIBS})
-target_link_libraries(fl_asr_test flashlight-app-asr ${CMAKE_DL_LIBS})
-target_link_libraries(fl_asr_decode flashlight-app-asr ${CMAKE_DL_LIBS})
+if (NOT FL_LIBRARIES_USE_MKL)
```
jacobkahn, line 65: Instead of linking this to binaries explicitly, the solution here is to make the `target_link_libraries` call for OpenMP `PUBLIC` so that targets that link to flashlight will have it transitively included automatically.

jacobkahn (changes requested):
We need a different approach. This is not how we want to do this.

We should link OpenMP only to the targets in flashlight that require it. This should fix the linker errors with all downstream tests and binaries. See my comment below - if OpenMP is actually required for binaries, and not just libraries, we should make the `target_link_libraries` call `PUBLIC` so that everything is propagated properly to downstream dependencies.

cmake/TestUtils.cmake
WilliamTambellini: sure, I tried to just target_link_libraries(OpenMP_CXX) but then nvcc was failing because not aware of the openmp flag. Retrying...

flashlight/app/asr/CMakeLists.txt
WilliamTambellini: fixed

cmake/TestUtils.cmake
WilliamTambellini: done

jacobkahn (approved):
There's probably a better way to do this, but this fixes my problem with gcc 7 and CMake 3.10. I'm sold.

Issue #270, opened 2020-11-19
Cuda sources (CriterionUtils.cu,...) compilation failure


WilliamTambellini:
### Bug Description
Cuda sources (CriterionUtils.cu,...) compilation failure

#### Reproduction Steps
- git pull the very master as today specially the upgrade to cmake 3.10 f5f84c464e63db83ed1bc876c968791bc9c5b20d
- cmake with 3.12+
- make

### Platform and Hardware
ubuntu 16, cuda 10.2

#### Additional Context
```
[ 12%] Building CUDA object CMakeFiles/fl-libraries.dir/flashlight/lib/sequence/criterion/cuda/CriterionUtils.cu.o
/usr/local/cuda/bin/nvcc  -DFL_LIBRARIES_USE_CUDA -DFL_LIBRARIES_USE_KENLM -DKENLM_MAX_ORDER=6 -I/home/wtambellini/repos/flashlight -I/usr/include/openblas -I/home/wtambellini/repos/kenlm -I/home/wtambellini/repos/flashlight/release/CUB/src/CUB -I/usr/local/cuda/include   -x cu -c /home/wtambellini/repos/flashlight/flashlight/lib/sequence/criterion/cuda/CriterionUtils.cu -o CMakeFiles/fl-libraries.dir/flashlight/lib/sequence/criterion/cuda/CriterionUtils.cu.o
/home/wtambellini/repos/flashlight/flashlight/lib/sequence/criterion/Defines.h(14): warning: scoped enumeration types are a C++11 feature
/home/wtambellini/repos/flashlight/flashlight/lib/sequence/criterion/cuda/CriterionUtils.cu(24): error: explicit type is missing ("int" assumed)
```
As you can see the std=c++11 is not set in the nvcc CLI.
It's because the cmake team has changed their mind several times about how to handle cuda.
Todays FL master is setting a global var in CUDAUtils.cmake: 
```
function (set_cuda_cxx_compile_flags)
  set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS};-std=c++11" PARENT_SCOPE)
  # Using host flags makes things bad - keep things clean
  set(CUDA_PROPAGATE_HOST_FLAGS OFF)
endfunction()
```
which does nt work anymore with 'recent' cmake.
 
The proper way to do is said to set target props:  adding 
```
  target_compile_options(fl-libraries PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--std=c++11>)
```
[...]

PR #294, opened 2020-12-03
Move nvcc setup to before any CUDA libs are added

jacobkahn:
Summary:
Fixes https://github.com/facebookresearch/flashlight/issues/270 with a good catch by WilliamTambellini -- we need to set `nvcc` flags before any calls to `add_library` since those freeze nvcc flags in place for those library. `-std=c++14` is now added to every nvcc invocation in fl given that setup and standard specification occurs earlier.

The existing code is technically a race between when global nvcc flags get set and when the libraries are added. This resolves that.

Differential Revision: D25304948


WilliamTambellini (commented):
Tks

PR #314, opened 2020-12-08
Many build and test fixes

jacobkahn:
Summary:
Fixes the following:
- Use GTest CMake integration — now, running `make test`, `ctest`, etc shows *all individual gtests* in CI and elsewhere!
- Fix a linking issue with `MemoryFrameworkTest` where, with the vcpkg install, `GTest::gmock` is a separate target and needs to explicitly be linked to tests for gmock symbols to be included.
- Adds `Findcub.cmake` to resolve an issue with `NOT ${CUB_INCLUDE_DIRS}` incorrectly being true even with the vcpkg build, which causes a re-download of CUB headers which conflict with the vcpkg Cub headers used to install ArrayFire
- Refactors how gtest targets work to merge targets found by CMake's native `FindGTest.cmake` -- which returns `GTest::GTest` and `GTest::Main`, both of which (should) include `gmock` with GTest >= 1.10.0. Coalesces around `GTest::gtest[_main]` and `GTest::gmock[_main]`.
- Resolves some flakiness with the bias gradient computation in `Autograd.LinearF16` test. Easter egg: does anyone actually read these diff summaries/know what any of this stuff means? If you see this note and mention it to me, you get a prize
- Changes `IMG_CLASS` in CMake options and build configurations to be `IMGCLASS` to be consistent with the namespace, directory, etc.
- Some readme fixes to vcpkg installation instructions, add `glog` as a dep
- Fixes an issue where `libgflags` was being added as an include directory for `flashlight-app-lm`
- Removes a use of `tmpnam`, which is [dangerous](https://www.gnu.org/software/libc/manual/html_node/Temporary-Files.html)

Differential Revision: D25409790


WilliamTambellini:
I do read them :)

PR #396, opened 2021-01-05
Multistage docker

alealv:
**Original Issue**: This dicussion started [here](https://github.com/facebookresearch/flashlight/pull/225#issuecomment-727635347)

### Summary
The main point of this PR is to update dockerfiles to use a multistage build, reducing build time and final image size. But it also adds to small but important features:

- `.dockerignore` to avoid passing unnecessary files to the build context
- `ccache` support to CMake

> I'm sure that most of the packages installed with apt on some builds and especially on the final image can be removed. But you know better which ones to take out. This would further reduce the final image size

| image                          | build-time | size   |
|--------------------------------|------------|--------|
| cpu-base-consolidation-latest  | 454.4s     | 4.76GB |
| cpu-latest                     | 433.6s     | 3.38GB |
| cuda-base-consolidation-latest | 970.3s     | 13.7GB |
| cuda-latest                    | 440.3s     | 9.31GB |

### DIFF
--- a/.docker/Dockerfile-CPU
+++ b/.docker/Dockerfile-CPU
@@ -4,18 +4,75 @@
 # flashlight       master     (git, CPU backend)
 # ==================================================================
 
-FROM flml/flashlight:cpu-base-consolidation-latest
-
-ENV MKLROOT="/opt/intel/mkl"
+FROM flml/flashlight:cpu-base-consolidation-latest as build
 
 # ==================================================================
 # flashlight with CPU backend
 # ------------------------------------------------------------------
 # Setup and build flashlight
-RUN mkdir /root/flashlight
+RUN mkdir /tmp/flashlight
 
-COPY . /root/flashlight
+COPY . /tmp/flashlight
 
-RUN cd /root/flashlight && mkdir -p build && \
-    cd build && cmake .. -DFL_BACKEND=CPU -DFL_LIBRARIES_USE_CUDA=OFF && \
-    make -j$(nproc) && make install
+RUN mkdir -p /tmp/flashlight/build && \
+    cd /tmp/flashlight/build && \ 
+    cmake -DCMAKE_BUILD_TYPE="Release" -DCMAKE_INSTALL_PREFIX="/opt/flashlight" -DFL_BACKEND="CPU" -DFL_LIBRARIES_USE_CUDA="OFF" -DFL_BUILD_TESTS="ON" -DDNNL_DIR="/opt/onednn/lib/cmake/dnnl" -DGloo_INCLUDE_DIR="/opt/gloo/include" -DGloo_NATIVE_LIBRARY="/opt/gloo/lib/libgloo.a" .. && \
+    make -j$(nproc) && \
+    make install -j$(nproc)
+
+
+
+#############################################################################
+#                            SECOND STAGE                                   #
+#############################################################################
+
+FROM ubuntu:18.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# install dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        g++ \
+        # for cmake
+        zlib1g-dev libcurl4-openssl-dev \
+        # for MKL
+        apt-transport-https \
+        gpg-agent gnupg2 \
+        # for arrayfire CPU backend
+        # OpenBLAS
+        libopenblas-dev libfftw3-dev liblapacke-dev \
+        # ATLAS
+        libatlas3-base libatlas-base-dev libfftw3-dev liblapacke-dev \
+        # ssh for OpenMPI
+        openssh-server openssh-client \
+        # OpenMPI
+        libopenmpi-dev libomp-dev openmpi-bin \
+        # libsndfile
+        libsndfile1-dev \
+        # for libsndfile for ubuntu 18.04
+        libopus-dev \
+        # FFTW
+        libfftw3-dev \
+        # for kenlm
+        zlib1g-dev libbz2-dev liblzma-dev \
+        # gflags
+        libgflags-dev libgflags2.2 \
+        # for glog
+        libgoogle-glog-dev libgoogle-glog0v5 \
+        # for receipts data processing
+        sox \
+        # for python
+        python3-dev python3-pip python3-distutils && \
+    apt-get clean && \
+    apt-get -y autoremove && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY --from=build /opt/arrayfire  /opt/arrayfire
+COPY --from=build /opt/kenlm      /opt/kenlm
+COPY --from=build /opt/intel      /opt/intel
+COPY --from=build /opt/boost      /opt/boost
+COPY --from=build /opt/flashlight /opt/flashlight
+
+ENV KENLM_ROOT_DIR=/opt/kenlm
+ENV MKLROOT="/opt/intel/mkl"
+ENV LD_LIBRARY_PATH="/opt/boost/lib:$LD_LIBRARY_PATH"
--- a/.docker/Dockerfile-CPU-Base
+++ b/.docker/Dockerfile-CPU-Base
@@ -17,7 +17,7 @@
 # boost            1.75.0       (source)
 # ==================================================================
 
-FROM ubuntu:18.04 as build
+FROM ubuntu:18.04 as cpu_base_builder
 
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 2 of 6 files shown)


tlikhomanenko:
Hey! Thanks for sending this PR!

We have now cuda 11, so I started two days ago to update the base images. I have tested the `--squash` option you suggested, it works fine. Images sizes for cuda base:
- without squash 16.1 GB
- with squash 11.8 GB
- with one RUN command 11.8 GB

So I think we are good to go with squash option for simplicity. What do you think @alealv?

tlikhomanenko:
I have the question: how did you get image for fl to be less than the base one?

alealv:
Because of the MultiStage Docker. The final image has the compiled flashlight files but not everything that's needed to compile.

 As I said, this could be also narrow down if we don't install system packages that are not needed for runtime, but I'm not sure which are those.

tlikhomanenko:
Let me play a bit with this and squash, but I would like to have the whole fl folder in docker as users could install the python bindings, tweak the code and recompile (I often do these sort of things).

tlikhomanenko:
I have a look at your pull request, docker multi-stage suggestion and its docs, and squash. Also had a thought on what we need/want at the end. Here is my summary, let me know what do you think @alealv @jacobkahn:
- Let's install fl and remove the build dir - thus we safe 2.5Gb of space + everyone can rebuild this fl commit with necessary fixes/extra code/extra params (prefer still to have this folder, to simplify life to have proper hash commit of fl)
- I see that for the final fl image @alealv you removed part of dependencies, for example installed vim/nano/emacs/python packages. We have added them in base image specifically to have base things in the image anyway + possibility to rebuild fl (with some own modifications)/build python bindings. I would say we don't need to have multi-stage for fl image (here we can simply just remove build dir to save the space).
- We are not really restricted by the image size, while still trying to keep minimalistic settings 
  - for example we remove python binding build because torch is not lightweight; 
  - do squash because it is simple to avoid multi-layer in the docker build and have reduced size.
- I have experimented with cpu base image build with squash or use your multi-stage proposal (I have modified a bit your version to have image with all apt installations as separate image)
  - multi stage image 4.14GB
  - squash image 3.96GB

Probably it is better to go multistage as soon as squash is experimental feature.

I will send some changes on top of your PR, so we can faster merge your contribution.

Also thanks for improving python, cmake installation. About boost - I prefer still have the version we had so only headers for arrayfire as we don't use boost anywhere else.

tlikhomanenko:
Here is PR for CPU docker on top of your. Let me know if you have any objections or strong preferences on what I have changed. Working now on CUDA docker. 

FYI: @jacobkahn refactored cmake for python bindings, and I am planning to work on `pip install` and putting on pypi bindings after docker PRs.

alealv:
> Let me play a bit with this and squash, but I would like to have the whole fl folder in docker as users could install the python bindings, tweak the code and recompile (I often do these sort of things).

If you take a closer look at the `.dockerignore`; I'm adding bindings folders
```sh
# Ignore everything
**

# Allow files and directories
!/bindings/**
!/cmake/**
!/flashlight/**
!/CMakeLists.txt
```
I exclude everything else because I thought it wasn't relevant for the final image (git uses a lot of space).

> Let's install fl and remove the build dir - thus we safe 2.5Gb of space + everyone can rebuild this fl commit with necessary fixes/extra code/extra params (prefer still to have this folder, to simplify life to have proper hash commit of fl)

Ok, then the second stages of `Docker-CPU` and `Docker-CUDA` are not needed.

> I see that for the final fl image @alealv you removed part of dependencies, for example installed vim/nano/emacs/python packages. We have added them in base image specifically to have base things in the image anyway + possibility to rebuild fl (with some own modifications)/build python bindings. I would say we don't need to have multi-stage for fl image (here we can simply just remove build dir to save the space).

Yes, my intention was to keep the minimum size possible. I guess it won't increase the size so much If you add these editors packages, although you could install them whenever you need to further work with the image.

> FYI: @jacobkahn refactored cmake for python bindings (it is landed), and I am planning to work on pip install and putting on pypi bindings after docker PRs.

Cool

> Working now on CUDA docker.

I thought that you were using Ubuntu 18.04 for a specific reason. If you are moving to Ubuntu 20.04, then I recommend using the 
[11.1-cudnn8-devel-ubuntu20.04](https://ngc.nvidia.com/catalog/containers/nvidia:cuda/tags) image (which was [recently uploaded](https://gitlab.com/nvidia/container-images/cuda/-/issues/83#note_462761385)) 


I also commented on your commit.

.docker/Dockerfile-CPU
```diff
@@ -4,18 +4,75 @@
...
 # ------------------------------------------------------------------
 # Setup and build flashlight
-RUN mkdir /root/flashlight
+RUN mkdir /tmp/flashlight
+
+COPY . /tmp/flashlight
+
+RUN mkdir -p /tmp/flashlight/build && \
+    cd /tmp/flashlight/build && \ 
+    cmake -DCMAKE_BUILD_TYPE="Release" -DCMAKE_INSTALL_PREFIX="/opt/flashlight" -DFL_BACKEND="CPU" -DFL_LIBRARIES_USE_CUDA="OFF" -DFL_BUILD_TESTS="ON" -DDNNL_DIR="/opt/onednn/lib/cmake/dnnl" -DGloo_INCLUDE_DIR="/opt/gloo/include" -DGloo_NATIVE_LIBRARY="/opt/gloo/lib/libgloo.a" .. && \
```
jacobkahn, line 19: `Gloo_INCLUDE_DIR` and `Gloo_NATIVE_LIBRARY` won't be used anymore. Using `-DGloo_DIR=/opt/gloo/lib/cmake` (or wherever the `gloo-targets.cmake` is) will be fine.

.docker/Dockerfile-CPU
```diff
@@ -4,18 +4,75 @@
...
+    make -j$(nproc) && \
```
jacobkahn, line 20:
```suggestion
```

.docker/Dockerfile-CPU
```diff
@@ -4,18 +4,75 @@
...
+        gpg-agent gnupg2 \
+        # for arrayfire CPU backend
+        # OpenBLAS
+        libopenblas-dev libfftw3-dev liblapacke-dev \
+        # ATLAS
+        libatlas3-base libatlas-base-dev libfftw3-dev liblapacke-dev \
+        # ssh for OpenMPI
+        openssh-server openssh-client \
+        # OpenMPI
+        libopenmpi-dev libomp-dev openmpi-bin \
```
jacobkahn, line 49: `libomp-dev` ≠ OpenMPI - it's the LLVM OpenMP runtime - and it could interact poorly with MKL's iomp which already fulfills that requirement

.docker/Dockerfile-CPU
```diff
@@ -4,18 +4,75 @@
...
+        # libsndfile
+        libsndfile1-dev \
+        # for libsndfile for ubuntu 18.04
+        libopus-dev \
```
jacobkahn, line 53: are we sure `libopus` is needed? I think the symbols `libsndfile1-dev` expects are already in `libogg` which is a transitive libsndfile dependency

.docker/Dockerfile-CPU-Base
```diff
@@ -3,38 +3,144 @@
...
-# MKL              2018.4-057   (apt)
-# arrayfire        3.7.1        (git, CPU backend)
-# oneDNN           2.0          (git)
-# Gloo             b7e0906      (git)
-# libsndfile       4bdd741      (git)
+# cmake            3.10         (oficial binaries)
+# MKL              2020.4-912   (apt)
+# arrayfire        3.7.3        (git, CPU backend)
+# libsndfile       latest       (apt, v1.0.28-4)
+# oneDNN           master       (git)
```
jacobkahn, line 10: let's freeze at `v2.0`, which is what we're using in CI

.docker/Dockerfile-CPU-Base
```diff
@@ -3,38 +3,144 @@
...
+        #apt-transport-https \
+        #gpg-agent gnupg2 \
+        # ssh for OpenMPI
+        #openssh-server openssh-client \
+        ## gflags
+        #libgflags-dev libgflags2.2 \
+        ## for glog
+        #libgoogle-glog-dev libgoogle-glog0v5 \
+
+# Install boost 1.75
```
jacobkahn, line 42: with a Ubuntu 20xx image, we can use `libboost-all-dev` which will be 1.71.0. Should be fine.

.docker/Dockerfile-CPU-Base
```diff
@@ -3,38 +3,144 @@
...
         build-essential \
         ca-certificates \
-        wget \
+        curl \
+        git \
+        g++
+        # for cmake
+        #zlib1g-dev libcurl4-openssl-dev \
+        # for MKL
+        #apt-transport-https \
```
jacobkahn, line 33: are we not installing MKL? I also see CBLAS install cmds below

.docker/Dockerfile-CPU-Base
```diff
@@ -3,38 +3,144 @@
...
+
+# cmake 3.10
+RUN curl -sSLo - https://cmake.org/files/v3.10/cmake-3.10.3-Linux-x86_64.tar.gz | tar -xz -C /opt/cmake --strip-components 1
+ENV PATH="/opt/cmake/bin:$PATH"
+# ==================================================================
+# arrayfire with CPU backend https://github.com/arrayfire/arrayfire/wiki/Build-Instructions-for-Linux
+# ------------------------------------------------------------------
+FROM cpu_base_builder as cpu_arrayfire
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
jacobkahn, line 59: Shouldn't be needed if installing MKL (we do need fftw3-dev)

.docker/Dockerfile-CPU-Base
```diff
@@ -3,38 +3,144 @@
...
+        libatlas3-base libatlas-base-dev liblapacke-dev \
+        # FFTW
+        libfftw3-dev
+
+# Install ArrayFire
+RUN cd /tmp && \
+    git clone --branch v3.7.3 --depth 1 --recursive --shallow-submodules https://github.com/arrayfire/arrayfire.git && \
+    mkdir -p arrayfire/build && \
+    cd arrayfire/build && \
+    cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/opt/arrayfire -DAF_BUILD_CUDA=OFF -DAF_BUILD_CPU=ON -DAF_BUILD_OPENCL=OFF -DAF_BUILD_EXAMPLES=OFF -DBUILD_TESTING=OFF -DAF_WITH_IMAGEIO=OFF .. && \
```
jacobkahn, line 72: add `-DAF_BUILD_DOCS=OFF`

.docker/Dockerfile-CPU-Base
```diff
@@ -3,38 +3,144 @@
...
+    make -j$(nproc) && \
+    make install -j$(nproc)
+
+# ==================================================================
+# oneDNN https://github.com/oneapi-src/oneDNN#requirements-for-building-from-source
+# ------------------------------------------------------------------
+FROM cpu_base_builder as cpu_onednn
+
+#Install OpenMP and TBB
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
jacobkahn, line 82: again, not needed if we're using MKL

.docker/Dockerfile-CPU-Base
```diff
@@ -3,38 +3,144 @@
...
+# Instal gzip, bz2 and xz
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        zlib1g-dev libbz2-dev liblzma-dev
+
+RUN cd /tmp && \
+    git clone --depth 1 https://github.com/kpu/kenlm.git && \
+    mkdir -p kenlm/build && \
+    cd kenlm/build && \
+    cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/opt/kenlm -DCMAKE_POSITION_INDEPENDENT_CODE=ON .. && \
+    make -j$(nproc) && \
```
jacobkahn, line 122:
```suggestion
```

CMakeLists.txt
```diff
@@ -6,6 +6,10 @@ include(CTest)
 include(CMakeDependentOption)
 
 # ----------------------------- Setup -----------------------------
+find_program(CCACHE_PROGRAM ccache)
```
jacobkahn, line 9: do you see better performance with ccache over incremental `make`? Just curious

jacobkahn (commented):
Thanks for the PR, @alealv.

I now see that @tlikhomanenko has resolved many of my comments in https://github.com/facebookresearch/flashlight/pull/401, so I'll make sure those updates can be made there. You should have a look at that PR too.

CMakeLists.txt
alealv: It's much faster if you continuously rebuild the project.

alealv:
I agree with all your comments. Although I think @tlikhomanenko is working on some improvements #401

I commented on the code but Github doesn't seem to map the comments to the corresponding lines.

tlikhomanenko:
Yep, let's continue discussion and changes in mine PR, so all contribution from @alealv is accounted and also we can faster move as I can simpler test/fix all comments on my own.

About Ubuntu - we set in docker the one with which we are working on our machines, recently we had upgrade to 20.04, so testing now CUDA images for 11 ubuntu 18 and 11.1 ubuntu 20 to pass all tests.

.docker/Dockerfile-CPU
tlikhomanenko: Let me recheck this, but when I tried on ubuntu 18 libopus was not in the list of dependencies and apt-packages which are installed with libsndfile1.

.docker/Dockerfile-CPU
tlikhomanenko: will be solving in https://github.com/facebookresearch/flashlight/pull/401

.docker/Dockerfile-CPU
tlikhomanenko: will be solving in https://github.com/facebookresearch/flashlight/pull/401

.docker/Dockerfile-CPU-Base
tlikhomanenko: done in https://github.com/facebookresearch/flashlight/pull/401

.docker/Dockerfile-CPU-Base
tlikhomanenko: done in https://github.com/facebookresearch/flashlight/pull/401 (for kenlm dep)

.docker/Dockerfile-CPU-Base
tlikhomanenko: we are not installing for image which used in some deps installation, but directly install in the final image. More clean version find in https://github.com/facebookresearch/flashlight/pull/401.

.docker/Dockerfile-CPU-Base
tlikhomanenko: will be solved in https://github.com/facebookresearch/flashlight/pull/401

.docker/Dockerfile-CPU
tlikhomanenko:
```
root@176d6b2a8ddb:/# apt-get install libsndfile1-dev
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
  libdpkg-perl libfile-fcntllock-perl libflac-dev libflac8 libgdbm-compat4 libgdbm6 libglib2.0-0 libglib2.0-data liblocale-gettext-perl libogg-dev libogg0 libperl5.30 libsndfile1 libvorbis-dev
  libvorbis0a libvorbisenc2 libvorbisfile3 netbase perl perl-modules-5.30 pkg-config shared-mime-info xdg-user-dirs xz-utils
```

tlikhomanenko:
Here is the final version on CUDA docker and all CI fixes with switching to new docker base images https://github.com/facebookresearch/flashlight/pull/407

.docker/Dockerfile-CPU-Base
alealv: Arrayfire is not needed?

.docker/Dockerfile-CPU-Base
alealv: but it's required for [OneDNN](https://github.com/oneapi-src/oneDNN#cpu-engine), or not?

tlikhomanenko:
@alealv could you rebase to recent master? I need the commit with python improvement and I think we are ready to land the current versions of docker. Thanks for all your great ideas and PR on improvement. I will revisit again on squeezing the docker image size as we solve some issues on af build with mkl.

alealv:
> @alealv could you rebase to recent master? I need the commit with python improvement and I think we are ready to land the current versions of docker. Thanks for all your great ideas and PR on improvement. I will revisit again on squeezing the docker image size as we solve some issues on af build with mkl.

Done. I'm glad to contribute!

PR #401, opened 2021-01-07
refactor of cpu docker

tlikhomanenko:
Summary:
Refactor CPU docker on top of PR https://github.com/facebookresearch/flashlight/pull/396

- revert back the fl cpu image as we want to rebuild/reinstall/build python bindings, so preserve all installations and apt deps for user simplicity
- final base image 4GB
- final fl image 4.9GB
- preserve build folder for simpler checking the tests, that everything is working fine (for user)
- change deps installation:
  - upgrade to ubuntu to 20.04
  - upgrade kenlm commit
  - remove boost, only use it for arrayfire as we had before
  - upgrade cmake to 3.16.3
- all tests passed with current fl

Differential Revision: D25821428

### DIFF
--- a/.docker/Dockerfile-CPU
+++ b/.docker/Dockerfile-CPU
@@ -4,75 +4,28 @@
 # flashlight       master     (git, CPU backend)
 # ==================================================================
 
-FROM flml/flashlight:cpu-base-consolidation-latest as build
+FROM flml/flashlight:cpu-base-consolidation-latest
+
+# just in case for visibility
+ENV PATH="/opt/cmake/bin:$PATH"
+ENV MKLROOT="/opt/intel/mkl"
+ENV KENLM_ROOT=/opt/kenlm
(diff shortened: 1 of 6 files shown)


alealv:
> I need this, because final image I didn't reinstall all these things, see line 114 `FROM cpu_base_builder as cpu_final`

Yes but you are cleaning everything with line 137-139 anyway, so I guess it's not necessary to do it twice.

tlikhomanenko:
> > I need this, because final image I didn't reinstall all these things, see line 114 `FROM cpu_base_builder as cpu_final`
> 
> True, but you are cleaning everything with line 137-139 anyway (also in lines 150-152), so I guess you only need to do it once, in the end.

I do twice to prevent additional space from several RUN commands (otherwise need to run with squash). For now assume that we don't use squash (as experimental feature).

PR #407, opened 2021-01-08
refactor of cuda docker

tlikhomanenko:
Summary:
Refactor CUDA docker on top of PR https://github.com/facebookresearch/flashlight/pull/396
- revert back the fl cuda image as we want to rebuild/reinstall/build python bindings, so preserve all installations and apt deps for user simplicity
- final base image 14.6GB
- final fl image 15.5GB
- preserve build folder for simpler checking the tests, that everything is working fine (for user)
- change deps installation:
  - upgrade to ubuntu to 20.04, CUDA 11.1
  - upgrade kenlm commit
  - remove boost source build

Update all CI, for now base images I pushed as tmp, before land will fix to the right names:
- ubuntu 20 has no gcc-5 anymore, so I remove it

Differential Revision: D25846151

### DIFF
--- a/.docker/Dockerfile-CUDA-Base
+++ b/.docker/Dockerfile-CUDA-Base
@@ -2,49 +2,48 @@
 # module list
 # ------------------------------------------------------------------
 # Ubuntu           18.04
-# CUDA             10.0
-# CuDNN            7-dev
-# cmake            3.10         (git)
-# arrayfire        3.7.1        (git, CUDA backend)
 # OpenMPI          latest       (apt)
-# libsndfile       4bdd741      (git)
-# MKL              2018.4.057   (apt)
+# CUDA             11.1
+# CuDNN            8-dev
+# cmake            3.10         (oficial binaries)
+# MKL              2020.4-912   (apt)
+# arrayfire        3.7.3        (git, CUDA backend)
+# libsndfile       latest       (apt, 1.0.28-4)
 # FFTW             latest       (apt)
 # KenLM            4a27753      (git)
 # GLOG             latest       (apt)
 # gflags           latest       (apt)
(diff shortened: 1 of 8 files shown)


.docker/Dockerfile-CUDA-Base
```diff
@@ -1,142 +1,149 @@
...
+        apt-transport-https gpg-agent gnupg2 \
+        # for arrayfire and kenlm
+        libboost-all-dev \
+        # for arrayfire CPU backend
+        # OpenBLAS
+        libopenblas-dev liblapacke-dev \
+        # ATLAS
+        libatlas3-base libatlas-base-dev liblapacke-dev \
+        # FFTW
+        libfftw3-dev \
```
alealv, line 48: Not needed with CUDA backend
```suggestion
```

.docker/Dockerfile-CUDA-Base
```diff
@@ -1,142 +1,149 @@
...
-        tqdm \
-        numpy
+        vim \
+        emacs \
+        nano \
+        htop \
+        # nccl: for flashlight
+        libnccl2 libnccl-dev libglfw3-dev \
+        # for libsndfile for ubuntu 18.04, 20.04
+        libopus-dev \
```
alealv, line 112: Do we need this? It's not listed [here](https://github.com/facebookresearch/flashlight/blob/master/README.md#dependencies)

alealv (commented):
I have only one question. Why are you building flashlight by getting inside the docker and running the build commands, instead of just run `docker build` outside?

Also, I would add the compilation option `-DFL_LIBRARIES_BUILD_FOR_PYTHON=ON` for building the python bindings as well

PR #549, opened 2021-04-19
CI: Use GCC 10 for building the Docker images

EwoutH:
Use GCC 10 for building the Docker images instead of the GitHub default GCC 9. GitHub environments are not super stable so also makes sure the major compiler version isn't changed on accident.

Also adds a few white lines for readability.

**Original Issue**: None yet, discussion can take place in this PR since the change is quite small and specific.

**Test Plan (required)**
As far as I understand Circle CI tests the docker images. More tests can be done if needed.

### DIFF
--- a/.github/workflows/docker_image_build.yml
+++ b/.github/workflows/docker_image_build.yml
@@ -3,11 +3,16 @@ on:
   push:
     branches:
       - master
+
 jobs:
   cuda_image_build:
     if: github.repository_owner == 'flashlight'
     name: CUDA image build
     runs-on: ubuntu-latest
+    env:
+       CC: gcc-10
+       CXX: g++-10
+
     steps:
     - uses: actions/checkout@master
     - name: Build the CUDA Docker image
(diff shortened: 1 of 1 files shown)


jacobkahn:
@EwoutH thanks for the PR - it turns out that we aren't actually using Github's env to build things - the `docker build ...` command in the Github action builds Flashlight inside of the base container, which is determined by the env in [`.docker/Dockerfile-CUDA-Base`](https://github.com/flashlight/flashlight/blob/master/.docker/Dockerfile-CUDA-Base).

That env has the default gcc/g++ from apt that ships in Ubuntu 20.04, which is gcc 9. So we're not at the mercy of Github's env after all! Is there a particular problem you were seeing that you think using gcc 10 fixes?

EwoutH:
You’re right, in that case this PR can be closed. I mainly wanted to prevent an unexpected change in compiler when GitHub updates their defaults.

jacobkahn:
@EwoutH makes sense — if you have other ideas for how we can better leverage Github Actions (code coverage, automatic packaging, etc), we'd welcome PRs! 👍

PR #564, opened 2021-04-27
[fl::Tensor][1/N] Compute: sync and eval

jacobkahn:
Adds `flashlight/fl/tensor`.

For now, add `fl::sync()` and `fl::eval(af::array&)` to begin abstracting away ArrayFire-specific semantics into a general API. Changes all callsites throughout all Flashlight components.

For now, the ArrayFire implementations of these functions will live in `flashlight/fl/tensor/*.cpp`. These will be moved into `flashlight/fl/tensor/backend/af/*` (or some similar path) once the interface is complete.

Test plan:
Rebuild + `make test` + CI


jacobkahn:
Will export from an internal diff.

PR #591, opened 2021-05-10
Start to use std::filesystem

xuqiantong:
Summary: Using c++17 feature from `std::filesystem` to create tmp folders.

Differential Revision: D28333939


jacobkahn:
Integrated elsewhere.

PR #625, opened 2021-06-05
[build] Use Kenlm imported targets

jacobkahn:
Awaiting https://github.com/kpu/kenlm/pull/342, which fixes include directories and makes using imported targets for KenLM in flashlight possible. Removes some unfortunate silliness with having to append `include/kenlm` to include paths to capture internal transitively-included kenlm headers, since these are now specified on the `INSTALL_INTERFACE` of exported KenLM targets in https://github.com/kpu/kenlm/pull/342 and the correct directory appears in the `INTERFACE_INCLUDE_DIRECTORIES` of `IMPORTED` targets.

Will require updating Docker images to a more recent KenLM commit to make CI pass.

### Test Plan
Local build with Kenlm:
```
cmake .. -DCMAKE_INSTALL_PREFIX=/tmp/kenlm_test -DCOMPILE_TESTS=OFF
make install -j80
```
from FL build:
```
cmake .. -DFL_BACKEND=CUDA -DFL_BUILD_TESTS=OFF -DCMAKE_BUILD_TYPE=Debug -DFL_BUILD_LIBRARIES=ON -Dkenlm_DIR=/tmp/kenlm_test/share/kenlm/cmake/
make -j80
```

### DIFF
--- a/flashlight/lib/text/decoder/lm/CMakeLists.txt
+++ b/flashlight/lib/text/decoder/lm/CMakeLists.txt
@@ -12,16 +12,18 @@ target_sources(
 # ------------------------- KenLM-specific -------------------------
 
 if (FL_LIBRARIES_USE_KENLM)
-  find_package(kenlm)
+  # Try to find the CMake config first
+  find_package(kenlm CONFIG)
   if (NOT kenlm_FOUND)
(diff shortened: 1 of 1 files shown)


jacobkahn:
Done via Flashlight Text.

PR #733, opened 2021-08-30
Checkout to specific commit in tutorials on RASR models

tlikhomanenko:
**Original Issue**: 

Conformer was changed in the recent commit 0029fc4fbe71f8a109f19687a1cab3db91163794 and thus rasr models can be used only before it.

### Summary
Add git checkout in the notebooks

### Test Plan (required)
Preserve state as before breaking commits.


ultrasaurus:
fixes https://github.com/flashlight/flashlight/issues/772

jacobkahn:
@ultrasaurus — thanks for the flag. We're changing a bunch of things related to the build, CI, branch defaults, and those colabs in the next few weeks; I'll make sure that happens.

Issue #745, opened 2021-10-03
Different log base for scoring calculation


fann1993814:
### Question
The log prob score in KenLM is caculated by log_10. (base is 10)
However, the decoder use log_e (base is e -> natural logarithm) for most case calculations, such as TrieLogAdd in Trie.cpp.
Is it make incorrect scores?

PR #746, opened 2021-10-03
Fix KenLM Scoring

fann1993814:
**Original Issue**:  Different log base for scoring calculation #745 

### Summary

KenLM score may need to rescale.
So, I Apply log formula of the change of base.
log e^Prob = log 10^Prob / log 10^e.

### Test Plan (required) 
Calculate score from inference code directly.


jacobkahn:
@fann1993814 thanks for the PR -- I think we need to make this configurable and default to no scaling. If you're able to add that/get our existing tests passing, we can think about landing.

Issue #986, opened 2022-08-18
Computing InputLen and `forwardSequentialModuleWithPadMask`


mmbejani:
### Bug Description
Do you want to train a network which contains `Transformer` or `Conformer`? If you put all thing correct, and build your model with `buildSequentialModule`, you may face with the following runtime error when call `forwardSequentialModuleWithPadMask`:
```
 [LayerNorm] Input size along the norm axis doesn't with axisSize.
```
This error raise because you change, `T`, (most of the times reducing) the length of input before feed it into `Transformer` module. 

#### How to resolve it!
Here is my idea to solve this problem (for branch `0.3`)
```
fl::Variable forwardSequentialModuleWithPadMask(
    const fl::Variable& input,
    std::shared_ptr<fl::Module> ntwrk,
    const af::array& inputSizes) {
  // expected input dims T x C x 1 x B
  int B = input.dims(3);
  auto inputMaxSize = af::tile(af::max(inputSizes), 1, B);
  af::array inputNotPaddedSizeRatio = inputSizes / inputMaxSize;
  auto ntwrkSeq = std::dynamic_pointer_cast<fl::Sequential>(ntwrk);
  auto output = input;
  for (auto& module : ntwrkSeq->modules()) {
    auto tr = std::dynamic_pointer_cast<fl::Transformer>(module);
    auto cfr = std::dynamic_pointer_cast<fl::Conformer>(module);
    if (tr != nullptr || cfr != nullptr) {
        /* input dims of Transformer module should be CxTxBx1 */
        int T = output.dims(1);
        auto inputNotPaddedSize = af::ceil(T * inputNotPaddedSizeRatio);
        auto padMask = af::iota(af::dim4(T, 1), af::dim4(1, B)) <
                                af::tile(inputNotPaddedSize, T, 1);
        output = module->forward({output, fl::noGrad(padMask)}).front();
    } else {
      output = module->forward({output}).front();
    }
  }
  return output.as(input.type());
}
```

mmbejani:
[https://github.com/flashlight/flashlight/pull/989](url)

PR #989, opened 2022-08-21
inference from a model with different types of layers based on forwar…

mmbejani:
…dSequentialModuleWithPadMask

### Summary
The function `forwardSequentialModuleWithPadMask` work for those networks which are built based on just `Transformer` family. If the network has non-`Transformer` based layer, e.g. Conv2D, then the length of the input sequence is changed however the length of the sequence is computed based initial length which raise the following runtime error:
```
 [LayerNorm] Input size along the norm axis doesn't with axisSize.
```

I compute the length of the sequence based on itself after applying each layer on the input.

### Test Plan (required)
Create networks with different combinations of modules `Transformers` and non-`Transformers` and applying the networks on random inputs.

### DIFF
--- a/flashlight/pkg/runtime/common/SequentialBuilder.cpp
+++ b/flashlight/pkg/runtime/common/SequentialBuilder.cpp
@@ -64,17 +64,20 @@ fl::Variable forwardSequentialModuleWithPadMask(
     std::shared_ptr<fl::Module> ntwrk,
     const af::array& inputSizes) {
   // expected input dims T x C x 1 x B
-  int T = input.dims(0), B = input.dims(3);
+  int B = input.dims(3);
   auto inputMaxSize = af::tile(af::max(inputSizes), 1, B);
-  af::array inputNotPaddedSize = af::ceil(inputSizes * T / inputMaxSize);
-  auto padMask = af::iota(af::dim4(T, 1), af::dim4(1, B)) <
-      af::tile(inputNotPaddedSize, T, 1);
+  af::array inputNotPaddedSizeRatio = inputSizes / inputMaxSize;
   auto ntwrkSeq = std::dynamic_pointer_cast<fl::Sequential>(ntwrk);
   auto output = input;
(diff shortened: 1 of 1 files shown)


mmbejani:
#986

flashlight/pkg/runtime/common/SequentialBuilder.cpp
```diff
@@ -64,17 +64,20 @@ fl::Variable forwardSequentialModuleWithPadMask(
...
-      af::tile(inputNotPaddedSize, T, 1);
+  af::array inputNotPaddedSizeRatio = inputSizes / inputMaxSize;
   auto ntwrkSeq = std::dynamic_pointer_cast<fl::Sequential>(ntwrk);
   auto output = input;
   for (auto& module : ntwrkSeq->modules()) {
     auto tr = std::dynamic_pointer_cast<fl::Transformer>(module);
     auto cfr = std::dynamic_pointer_cast<fl::Conformer>(module);
     if (tr != nullptr || cfr != nullptr) {
+      /* input dims of Transformer module should be CxTxBx1 */
+      int T = output.dims(1);
```
jacobkahn, line 77: As mentioned -- this is a big assumption that's architecture and layer dependent. Better to resize the padding mask per layer to fit whatever layers are in the architecture or add some abstraction to automatically do this if interested.

jacobkahn (changes requested):
Thanks for the PR! This approach unfortunately isn't general. It assumes that `T` for a given layer is in the 1st indexed dimension, which is architecture dependent.

The idea with `forwardWithPadMask` is that the padding will be computed for a given sequential, but that the user will resize the mask for each layer depending on its input size and expected dimensions.

You could reimplement this function in your user-space code, or you might consider creating a module that does this resizing for you in a way that you'd expect given a wrapped module (e.g. Conv2D). If you're able to think up a way to do this in a clean way, I'm happy to take a look at an updated PR.

PR #991, opened 2022-08-22
[gh-actions] Action for running lightweight oneDNN benchmarks

jacobkahn:
See title — action stub/starting point. Builds on Ubuntu 20.04.

We'll ship this lib to downstream projects pulling in `libflashlight` on install.

Test Plan: CI


jacobkahn:
Closing because we can't yet easily get beefy CPU machines on GH Actions without a lot of other setup. Will swap over to CircleCI for this for now.

PR #1010, opened 2022-09-26
Rename rename acoustic-model-specific callbacks

jacobkahn:
Summary: Rename speech-specific things (mostly `acoustic model`) to refer to an "emitting model" which might not be a speech-based model.

This is per https://github.com/flashlight/text/pull/18

Differential Revision: D39797755


jacobkahn (commented):
Update Flashlight Text base hash

PR #1017, opened 2022-10-03
[JitBackend] backend directory/cmake setup & basic IR

StrongerXi:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.** Changes *must* be discussed.

**Original Issue**: [corresponding issue on Github]

*Note:* You can add `closes #[issue number]` to automatically close the issue that this PR resolves when it is merged.

### Summary
As title.

This PR sets up 
- JIT IR's `Node` interface
- Some basic nodes
  - ScalarNode (representing a tensor filled with constant)
  - ValueNode (representing a tensor -- useful for input and lazy eval)
  - BinaryNode (representing binary tensor operations)
  - CustomNode (representing operation with customized evaluation logic -- useful for backend-specific optimization like node fusion, or as a fallback node to capture maximal graph)

### Test Plan (required)
Unit tests are added.


flashlight/fl/tensor/backend/jit/ir/Node.h
```diff
@@ -0,0 +1,118 @@
...
+  const std::vector<Node*>& inputs() const;
+  void setInput(unsigned inputIdx, Node* newInput);
+
+  // Uses
+  const UseList& uses() const;
+  void replaceAllUsesWith(Node* newInput);
+
+  // Mainly for debugging/testing
+  unsigned getRefCount() const;
+  // Use carefully -- this manually simulates shared ownership of a node
```
bwasti, line 72: why not use `std::shared_ptr`?

flashlight/fl/tensor/backend/jit/ir/Node.cpp
```diff
@@ -0,0 +1,135 @@
...
+  // reset
+  inputUseIters_[inputIdx] = oldInput->uses_.end();
+  inputs_[inputIdx] = nullptr;
+  // clean up metadata
+  delete *oldInputUseIter;
+  oldInput->uses_.erase(oldInputUseIter);
+  oldInput->decRefCount();
+}
+
+Node::Node(std::vector<Node*>&& inputs) {
```
bwasti, line 48: you may want to calculate a quick `hash` here :)

flashlight/fl/tensor/backend/jit/ir/NodeType.h
```diff
@@ -0,0 +1,36 @@
...
+  Scalar,
+  Value,
+};
+
+/**
+ * Return a readable string representation of the given node type.
+ *
+ * @return a string that represents the given node type.
+ */
+std::string nodeTypeToString(const NodeType type);
```
jacobkahn, line 29: `#include <string>`

flashlight/fl/tensor/backend/jit/ir/ScalarNode.h
```diff
@@ -0,0 +1,76 @@
...
+        return new ScalarNode(shape, type, static_cast<long long>(scalar));
+      case dtype::u64:
+        return new ScalarNode(
+            shape, type, static_cast<unsigned long long>(scalar));
+      case dtype::f16:
+      case dtype::f32:
+      case dtype::f64:
+        return new ScalarNode(shape, type, static_cast<double>(scalar));
+    }
+    throw std::runtime_error("Unknown dtype");
```
jacobkahn, line 55: Nit -- more descriptive exception message

flashlight/fl/tensor/backend/jit/ir/ScalarNode.h
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  const Shape& shape() const;
+  dtype dataType() const;
+
+  // cast to T
+  template <typename T>
+  T scalar() const {
+    // TODO once we have full c++17 support, use `std::visit(overloaded {...})`
```
jacobkahn, line 64: We actually fully support C++17 at this point, so feel free to use `std::visit`

You probably don't need an overloaded template trick though: you can probably actually just do something like
```cpp
return std::visit([](auto&& val) { return val; }, scalar_);
```

jacobkahn (changes requested):
👏 👏 👏 

Super exciting!! A few quick points, then let's get this landed!

flashlight/fl/tensor/backend/jit/ir/Node.h
StrongerXi: Yeah I've been struggling with this thing for a long time. Let me document my thought process here:

## Lifetime Management Goals

So my goals for lifetime management here are
1. safe (no memory leak)
2. transparent to the client (as if we have GC)
3. fast (JIT needs to be fast)

## Ownership Requirements
Shared ownership is inevitable in this JIT. There are already these clients that hold nodes for variable amount of time:
1. A `JitTensor` holds a node
2. A `Node` may hold other input nodes, and user nodes.
3. A `GraphVizPrinter` may hold a node for some time, so we can print out not just graphs one by one independently, but also be able to show "shared subgraphs" -- this is particularly helpful for debugging.

## Ownership Implementations
At the very beginning I considered (1). multiple owners (shared_ptr, refcounting, etc.) v.s. (2). a single long-living ownership (like a `Graph` class that holds all nodes).

### Single long-living owner
[PyTorch's JIT](https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/ir/ir.h#L92-L95) uses (2), but I kinda don't see any other reason to introduce a `Graph` class, and I _think_ it still requires some ["manual delete"](https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/ir/ir.h#L726-L734). So I ditched that idea... Maybe I should look more into that direction?

### Multiple Owners

#### std::shared_ptr
I started with `std::shared_ptr`, ran into issues when I tried to represent `Node-->UserNode` relations, because it creates a cycle with the `Node-->InputNode` relations, and the latter cannot be avoided. However, with the `Use` abstraction now, this actually becomes a nonissue.

So the only concern here is **performance** . But that's just **speculation**, I never measured it.

#### Manual refcount
Basically
1. Restrict ownership sharing to the 3 "big" owners (listed above), i.e., temporary users in places like optimization pass don't share ownership.
2. Manually control shared ownership by using `Node::incRefCount` and `Node::decRefCount`

## Conclusion
1. `std::shared_ptr`
  - __might__ hurt performance
  - slightly more verbose.
2. manual refcount 
  - fast
  - not completely automated

I'm leaning towards (2), and note that for (2), only the shared owners (see above) need to be careful. Ideally future optimization pass writers don't have to worry about this at all -- they just
```c++
oldNode->replaceAllUsesWith(CustomNode::create(...));
```

flashlight/fl/tensor/backend/jit/ir/Node.cpp
StrongerXi: Good point. I was thinking about that, but decided to keep it simple for now -- until hashing becomes needed later (e.g., for caching). Nice thing is that it shouldn't be hard to add.

flashlight/fl/tensor/backend/jit/ir/BinaryNode.h
```diff
@@ -0,0 +1,42 @@
...
+  static constexpr unsigned kLhsIdx = 0;
+  static constexpr unsigned kRhsIdx = 1;
+
+  // intentionally kept private
+  BinaryNode(Node* lhs, Node* rhs, BinaryOp op);
+
+ public:
+  static constexpr NodeType nodeType = NodeType::Binary;
+
+  static BinaryNode* create(Node* lhs, Node* rhs, BinaryOp op);
```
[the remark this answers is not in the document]
StrongerXi: Yes. [LLVM](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/IR/Instructions.h#L933-L965) kinda does this, but they have some customized `new` operator. Not sure what that is, but controlling allocation might allow us to do similar things in the future.

flashlight/fl/tensor/backend/jit/ir/Node.h
jacobkahn: I like this analysis. My thoughts:
1. Agree that an entire `Graph` notion is overkill, and ultimately constrains how different implementers might think about lifetime management
2. `shared_ptr` is annoying for the reasons you mention, and for another: contention with threading for both reads and writes to refcounts, even in places where thread safety might not matter, can be a big problem

The general philosophy with Flashlight and the `fl::Tensor` API thus far has been to automate less, aim for configuration over convention, and prioritize performance and simplicity, so this sounds very in line with that 😄

flashlight/fl/test/CMakeLists.txt
```diff
@@ -51,3 +51,4 @@ if (FL_USE_ONEDNN)
   build_test(SRC ${DIR}/tensor/onednn/OneDnnCPUStreamTest.cpp LIBS ${LIBS})
   build_test(SRC ${DIR}/tensor/onednn/OneDnnTensorTest.cpp LIBS ${LIBS})
 endif ()
+build_test(SRC ${DIR}/tensor/jit/JitNodeTest.cpp LIBS ${LIBS})
```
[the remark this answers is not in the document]
StrongerXi: Turned out to be a dumb "forgot to remove an old include" on my part.

The JIT IR is totally agnostic to other backends, so no need to gate it:).

StrongerXi:
- Address all feedbacks.
- Fix CI failures.
- Add more documentation for refcount ownership (see `Node::(inc/dec)RefCount`)

I'll do a bit more experiments with viability of this IR design (for JIT use) before publishing the PR.

flashlight/fl/tensor/backend/jit/ir/Node.h
```diff
@@ -0,0 +1,125 @@
...
+  //
+  // Possible stages of lifetime
```
StrongerXi, line 72: Hopefully this makes the refcount & ownership semantics clear.

Essentially, the real owners of a `Node` are the ones who invokes the `incRefCount` method (user Node, JitTensor, GraphVizPrinter). Again, the graph pass writers (the extensible part of the JIT) shouldn't have to deal with this at all.

jacobkahn (approved):
🎉  Looks great to me! Any other changes we can make incrementally once pushed.

PR #1020, opened 2022-10-05
Jit tensor full

StrongerXi:
Depends on #1019.

### Summary

"Parse" `full` in JIT:).

### Test Plan (required)
Unit tests are added.

```
make JitTensorTest
```


flashlight/fl/tensor/backend/jit/JitBackend.cpp
```diff
@@ -0,0 +1,537 @@
...
+    throw std::invalid_argument(                                              \
+        "JitBackend::fromScalar - not implemented for type " +                \
+        std::string(#TYPE));                                                  \
+  }                                                                           \
+  Tensor JitBackend::full(const Shape& shape, TYPE value, const dtype type) { \
+    switch (type) {                                                           \
+      case dtype::f16:                                                        \
+        return fullWithType(shape, value, dtype::f32).astype(dtype::f16);     \
+      case dtype::f32:                                                        \
+        return fullWithType(shape, value, type);                              \
```
jacobkahn, line 95: Can this function just be
```cpp
if (type == dtype::f16) {
  return fullWithType(shape, value, type).astype(dtype::f16(;
} else {
  return fullWithType(shape, value, type);
}
```
?

Actually: why is it necessary to have this cast with an fp16 type? Can't we just call `fullWithType` with the correctly-propagated value and let the underlying implementation handle the cast?

flashlight/fl/test/tensor/jit/JitNodeTest.cpp
```diff
@@ -129,6 +101,8 @@ TEST(JitNodeTest, getSetResult) {
   ASSERT_TRUE(node->getResult().has_value());
   ASSERT_TRUE(allClose(node->getResult().value(), tensor));
   ASSERT_THROW(node->setResult(tensor.copy()), std::invalid_argument);
+  // "free" node
```
jacobkahn, line 104:
```suggestion
// node is manually heap allocated
```

flashlight/fl/test/tensor/jit/JitTensorTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+  const auto node = jitTensor.node();
+  ASSERT_EQ(node->inputs(), NodeList({}));
+  ASSERT_EQ(node->getRefCount(), 1);
+  ASSERT_EQ(node->uses(), UseValList({}));
+  ASSERT_TRUE(node->isValue());
+  ASSERT_TRUE(allClose(data, node->getResult().value()));
+}
+
+TEST(JitTensorTest, full) {
+  fl::setDefaultTensorType<JitTensor<ArrayFireTensor>>();
```
jacobkahn, line 38: Is it worth having a test fixture that does this since this is happening for every test?

jacobkahn (commented):
Tried to keep comments to only the new things here, but I'll look again after a rebase

flashlight/fl/test/tensor/jit/JitNodeTest.cpp
StrongerXi: Actually they are all heap allocated, but these ones didn't transition to the automated shared ownership, so they must be deleted manually. Documented.

flashlight/fl/tensor/backend/jit/JitBackend.cpp
StrongerXi: My head was not right when I wrote this...

Yes and I was copying OneDnnBackend (and we couldn't do `fullWithType<float16>`) -- but JIT shouldn't worry about it.

Simplified.

StrongerXi:
Address all feedbacks.

StrongerXi:
Rebase.

jacobkahn (approved):
Awesome, thanks for those fixes

PR #1022, opened 2022-10-06
[JitBackend] IR evaluator

StrongerXi:
### Summary

We need to materialize the JIT graph when needed, e.g., (if client calls `tensor.toString()`).

Conceptually, `Evaluator`
1. does a **pass** through the JIT tree ("tree" since materialization is initiated from a single root node)
2. **dispatches** to another `TensorBackend`'s ops, e.g., `ScalarNode` gets dispatched to `backend.full(...)` 
3. **annotate** node with tensor result (only if necessary, to save memory footprint).

The JIT graph is untouched, leaving room for
1. optimizations starting from other root node that involves a shared subtree
4. node graph printing with performance hotspots.

The Evaluator only depends on the JIT IR and the Tensor API -- it's agnostic to the JitTensor (IR parser).

### Test Plan
Unit tests are added.

```
make JitEvaluatorTest JitTensorTest
./flashlight/fl/test/JitEvaluatorTest
```


flashlight/fl/test/tensor/jit/JitEvaluatorTest.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+TEST(JitEvaluatorTest, evalBinaryNode) {
+  // c1  c2
+  //  \  /
+  //   add
+  Shape shape(Shape({2, 2}));
+  auto dtype = dtype::s32;
+  const auto c1 = ScalarNode::create(shape, dtype, 1);
+  const auto c2 = ScalarNode::create(shape, dtype, 2);
+  const auto add = BinaryNode::create(c1, c2, BinaryOp::Add);
```
StrongerXi, line 53: Note that I intentionally avoided using `JitTensor` to build up the IR here -- to keep the boundary between Evaluator and JitTensor (which again, serves as the parser).

StrongerXi:
Rebase and fixing a dumb bug.

StrongerXi:
> Rebase and fixing a dumb bug.

Let me add a regression test for this bug -- basically a simple integration test for IR build-up and evaluation, like `(a + b).toString()`

StrongerXi:
Rebase.

flashlight/fl/tensor/backend/jit/JitBackend.cpp
```diff
@@ -31,10 +32,11 @@ TensorBackendType JitBackend::backendType() const {
...
 /* -------------------------- Compute Functions -------------------------- */
 
-void JitBackend::eval(const Tensor& /* tensor */) {
-  // Launch computation for a given tensor. Can be a noop for non-async
-  // runtimes.
-  FL_JIT_BACKEND_UNIMPLEMENTED;
+void JitBackend::eval(const Tensor& tensor) {
+  auto& jitTensor = toJitTensorBase(tensor);
+  assert(&jitTensor.backend() == this);
+  jitTensor.eval();
```
jacobkahn, line 38: This is interesting -- the separation of concerns here is a bit weird. Should the JIT wrap that call to eval for the most part and basically emit ops that another backend has to execute?

I can see a scenario in which the "deep" eval ends up doing excess computation, but maybe I'm thinking about this wrong

thoughts?

flashlight/fl/test/tensor/jit/JitTensorTest.cpp
```diff
@@ -102,6 +103,32 @@ TEST_F(JitTensorTest, div) {
...
+  Shape shape(Shape({2, 2}));
+  auto dtype = dtype::s32;
+  const auto t0 = full(shape, 11, dtype);
+  const auto t1 = full(shape, 22, dtype);
+  auto sum = t0 + t1;
+  ASSERT_FALSE(toJitTensorBase(sum).node()->getResult().has_value());
+  fl::eval(sum); // explicit eval call
+  ASSERT_TRUE(allClose(
+      toJitTensorBase(sum).node()->getResult().value(),
+      ArrayFireBackend::getInstance().full(shape, 33, dtype)));
```
jacobkahn, line 116: ditto with `DefaultTensorBackend.h` if we wanted to make this general

flashlight/fl/tensor/backend/jit/eval/Evaluator.h
```diff
@@ -0,0 +1,62 @@
...
+#include "flashlight/fl/tensor/backend/jit/ir/CustomNode.h"
+#include "flashlight/fl/tensor/backend/jit/ir/ScalarNode.h"
+
+namespace fl {
+
+/**
+ * A JIT tree evaluator. It dispatches to another Tensor Backend for carrying
+ * out the computation represented by the JIT tree.
+ */
+class Evaluator {
```
jacobkahn, line 24: Random idea, but is there any reason we might want to subclass this?

flashlight/fl/tensor/backend/jit/eval/Evaluator.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include "flashlight/fl/tensor/backend/jit/eval/Evaluator.h"
+
+#include "flashlight/fl/tensor/backend/jit/JitTensorBase.h"
+#include "flashlight/fl/tensor/backend/jit/ir/ValueNode.h"
+
+namespace fl {
+
+namespace {
+
+void fillNodeRefCountInTree(
```
jacobkahn, line 17: nit: document these?

flashlight/fl/tensor/backend/jit/eval/Evaluator.cpp
```diff
@@ -0,0 +1,132 @@
...
+    Node* root,
+    std::unordered_map<Node*, unsigned>& nodeToRefCount) {
+  if (nodeToRefCount.find(root) == nodeToRefCount.end()) {
+    nodeToRefCount.emplace(root, root->getRefCount());
+    for (const auto& input : root->inputs()) {
+      fillNodeRefCountInTree(input, nodeToRefCount);
```
jacobkahn, line 23: I'd sort of prefer this were iterative in preparation for having huge graphs, but either way

flashlight/fl/tensor/backend/jit/eval/Evaluator.cpp
```diff
@@ -0,0 +1,132 @@
...
+    case dtype::f64:
+      return backend_.full(shape, node.scalar<double>(), dtype);
+    case dtype::b8:
+    case dtype::s16:
+    case dtype::s32:
+    case dtype::s64:
+    case dtype::u8:
+    case dtype::u16:
+    case dtype::u32:
+      return backend_.full(shape, node.scalar<long long>(), dtype);
```
jacobkahn, line 89: Some of these widen pretty massively -- is that just to simplify the implementation?

jacobkahn (changes requested):

flashlight/fl/tensor/backend/jit/eval/Evaluator.h
StrongerXi: Can't really think of one yet -- I've had thoughts on supporting backend-specific Opaque node (from optimization, autograd, etc.), which requires the backend to inject evaluation logic for those nodes, which can be implemented either by inheritance or some TensorExtension-style thing.

flashlight/fl/tensor/backend/jit/eval/Evaluator.cpp
StrongerXi: Good point. I had this thought yesterday but forgot to update it.

flashlight/fl/tensor/backend/jit/eval/Evaluator.cpp
StrongerXi: > just to simplify the implementation?

Yes. I could use a variant that holds each possible type, but since `long long`, `unsigned long long` and `double` could capture all types w/o loss of precision, I went with these 3.

jacobkahn: Sure -- I guess this only matters for the memory footprint of `Node` and not an actual tensor...

flashlight/fl/tensor/backend/jit/JitBackend.cpp
StrongerXi: > the separation of concerns here is a bit weird

Agree. So the whole JitTensor backend serves as both **parser** and **driver** (mostly the former, but given the Tensor API, it also needs to be the one that kicks off the evaluation, e.g., upon `tensor.toString()` call.

> Should the JIT wrap that call to eval for the most part and basically emit ops that another backend has to execute?

Not sure if I fully understand the question. But `emit ops` sounds like the process of building up the JIT nodes/graph? I decided to put the `jitTensor.eval()`logic (launching Evaluator) inside `JitTensorBase`, because the it serves as the driver logic that'll call optimizer and update the node as well (in upcoming PR). It's slightly more indirect to update the node field from `JitBackend`.

> I can see a scenario in which the "deep" eval ends up doing excess computation, but maybe I'm thinking about this wrong

So the motivation for this was that for things like `JitTensor<JitTensor<T>>`, a single `fl::eval` won't kick off computation in `T`. But I don't have any other example in mind and this is unlikely to be used in practice:).

I guess it really depends on the desired semantics for `fl::eval` for Tensor clients, let me know what you think.

flashlight/fl/test/tensor/jit/JitEvaluatorTest.cpp
```diff
@@ -0,0 +1,124 @@
...
+ */
+
+#include <functional>
+
+#include <gtest/gtest.h>
+
+#include "flashlight/fl/tensor/Init.h"
+#include "flashlight/fl/tensor/Random.h"
+#include "flashlight/fl/tensor/Shape.h"
+#include "flashlight/fl/tensor/backend/af/ArrayFireBackend.h"
```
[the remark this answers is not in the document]
StrongerXi: Slight complication, `defaultTensorBackend()` returns the __current__ [dynamic](https://github.com/flashlight/flashlight/blob/main/flashlight/fl/tensor/DefaultTensorType.cpp#L14) default tensor backend, but `DefaultTensorBackend_t` is the static one, which we want here. 

I'll use `DefaultTensorBackend_t::getInstance()` for the moment, which is still better than explicitly calling AF backend.

Updated for the other test too, to remove dependency to explicit backend.

StrongerXi:
Address all feedbacks.

StrongerXi:
Leverage gtest fixture a bit more to centralize `DefaultTensorBackend_t::getInstance()` call.

flashlight/fl/tensor/backend/jit/JitBackend.cpp
jacobkahn: Yep, forgot it was the **driver** idea; vis-a-vis this abstraction the JIT would be constructing a modified graph vis a vis the wrapped backend. Makes perfect sense 👍

PR #1027, opened 2022-10-09
Fix jit memory leaks

StrongerXi:
Depends on #1022 (not a hard dependency; just the way the PR stack is)

### Summary
This PR
- fixes some memory leaks I discovered while running leak checking tools on JIT unit tests.
- simplifies `Node::replaceAllUsesWith`

Manual refcount is dangerous, but I think it'll pay off for future optimization pass writers.

### Test Plan (required)
Added unit tests, and ran them with `leaks` tool on MacOS.

```
make JitNodeTest JitTensorTest
./flashlight/fl/test/JitNodeTest
```


StrongerXi:
Accidentally closed this... Reopen.

PR #1028, opened 2022-10-09
Jit tensor optimizer

StrongerXi:
### Summary
The main motivation for the JIT component was optimization passes. Here it comes:
- __Optimizer__ is the abstraction for the optimization engine (trivial now, but might do more later -- caching, pass ordering, etc.).
- __Pass__ is the abstraction for optimization/transformation over JIT IR.
- __ScalarFoldingPass__ is a simple "Hello World" pass.

### Test Plan (required)
Unit tests are added
```
make JitScalarFoldingTest
```


flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
```diff
@@ -23,6 +23,22 @@ JitTensorBase::~JitTensorBase() {
   node_->decRefCount();
 }
 
+// take care of refcount for old & new nodes
+void JitTensorBase::replaceNode(Node* newNode) const {
+  if (node_ != newNode) {
+    node_->decRefCount();
+    newNode->incRefCount();
+    const_cast<JitTensorBase*>(this)->node_ = newNode;
```
StrongerXi, line 31: Welp. I think the root cause of this is that `TensorBackend::eval(const Tensor& tensor)`. But upon materialization I want to use optimizer to update JitTensor's `node` here.

I think the `const` in `TensorBackend::eval` really applies to the "Tensor evaluation result" (which is indeed unchanged here). But C++ `const` doesn't give us enough granularity to specify that...

jacobkahn: @StrongerXi that makes a lot of sense. Many methods in `TensorAdapter` are already intentionally non-const for this reason (because even inspection might modify internal state) -- feel free to send a separate change to make this non-const.

cc @bwasti for an example of something that we want `const` in the public API but don't want `const` internally

StrongerXi:
Rebase.

flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
```diff
@@ -23,6 +23,15 @@ JitTensorBase::~JitTensorBase() {
...
```
jacobkahn, line 31: Should this really be const/is it called from `const` members on `JitTensorBase`?

flashlight/fl/tensor/backend/jit/opt/Optimizer.cpp
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#include "flashlight/fl/tensor/backend/jit/opt/Optimizer.h"
+
+#include "flashlight/fl/tensor/backend/jit/opt/passes/ScalarFolding.h"
+
+namespace fl {
+
+Optimizer::Optimizer() {
+  passes_.emplace_back(std::make_unique<ScalarFolding>());
```
jacobkahn, line 15: I assume this will be made more configurable later, right? i.e. this isn't a default? 😄

flashlight/fl/tensor/backend/jit/opt/passes/ScalarFolding.h
```diff
@@ -0,0 +1,23 @@
...
+
+namespace fl {
+
+/**
+ * An optimization pass that recursively merges scalars inside a JIT tree,
+ * e.g., 1 + 2 becomes 3.
+ */
+class ScalarFolding : public Pass {
+ public:
+  Node* apply(Node* node);
```
jacobkahn, line 20: `override` needed?

flashlight/fl/tensor/backend/jit/opt/passes/ScalarFolding.cpp
```diff
@@ -0,0 +1,115 @@
...
+      node->replaceAllUsesWith(foldedScalar);
+      return foldedScalar;
+    }
+  }
+  return node;
+}
+
+Node* foldScalars(Node* node) {
+  for (const auto& input : node->inputs()) {
+    foldScalars(input);
```
jacobkahn, line 96: Just curious -- why apply this optimization in a prefix way? Makes sense that you're folding all of the inputs, but what if you wanted to apply other optimizations to the inputs?

flashlight/fl/tensor/backend/jit/opt/passes/ScalarFolding.cpp
```diff
@@ -0,0 +1,115 @@
...
+  T resVal = foldScalars(lhsVal, rhsVal, op);
+  return ScalarNode::create(Shape(lhs.shape()), type, resVal);
+}
+
+ScalarNode* foldScalarNodes(
+    const ScalarNode& lhs,
+    const ScalarNode& rhs,
+    const BinaryOp op,
+    const dtype type) {
+  switch (type) {
```
jacobkahn, line 49: No better idea at the moment, but I wish there were a better way to trait away what will invariably be many runtime --> compile time type switches

jacobkahn (approved):
Nice! A few minor things/questions

flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
StrongerXi: Yes, the call chain is

`JitTensorBackend::eval --> JitTensorBase::eval --> JitTensorBase::replaceNode`.

As discussed in earlier comment, the best solution would be making `TensorBackend::eval` take in non-const `Tensor`. I'll send another PR soon.

flashlight/fl/tensor/backend/jit/opt/Optimizer.cpp
StrongerXi: Yep. Just haven't figured out the best configuration API yet. Everything in the Optimizer and Evaluator are meant to be minimal but extensible atm.

Dropping a TODO.

flashlight/fl/tensor/backend/jit/opt/passes/ScalarFolding.cpp
StrongerXi: > what if you wanted to apply other optimizations to the inputs?

Wouldn't that be another pass, since this is just a scalar-folding pass?

So it's a prefix-traversal because we want to reduce nested scalar-binops in 1 single pass, e.g.,

```
2   3
 \  /
  *   4
   \ /
    +
```

flashlight/fl/tensor/backend/jit/opt/passes/ScalarFolding.cpp
StrongerXi: Yeah I've been observing and thinking about this.

I have a macro-based idea in mind, and I'll send a proof-of-concept PR at some point.

StrongerXi:
1. Address all feedbacks.
2. Rebase
3. Don't throw on `f16`, just skip it.

StrongerXi:
Better English.

flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
jacobkahn: Sounds good, although it's probably a lot better if `fl::eval` still took a const Tensor — `const_cast` seems a bit ugly. Maybe we can discuss on the PR — thanks for looking into that!

flashlight/fl/tensor/backend/jit/opt/passes/ScalarFolding.cpp
jacobkahn: 👍

jacobkahn (approved):

PR #1029, opened 2022-10-09
[JitBackend] Fix memory leaks in `Node::replaceAllUsesWith` and unit tests

StrongerXi:
### Summary
This PR
- fixes some memory leaks I discovered while running leak checking tools on JIT unit tests.
- simplifies `Node::replaceAllUsesWith`

Manual refcount is dangerous, but I think it'll pay off for future optimization pass writers.

### Test Plan (required)
Added unit tests, and ran them with `leaks` tool on MacOS.

```
make JitNodeTest JitTensorTest
./flashlight/fl/test/JitNodeTest
```

### DIFF
--- a/flashlight/fl/tensor/backend/jit/ir/Node.cpp
+++ b/flashlight/fl/tensor/backend/jit/ir/Node.cpp
@@ -77,13 +77,11 @@ const UseList& Node::uses() const {
 }
 
 void Node::replaceAllUsesWith(Node* newInput) {
-  refCount_++; // a trick to avoid freeing this node in last iteration
-  while (!uses_.empty()) { // each iteration updates uses
+  // each iteration updates links an existing user to newInput
+  while (!uses_.empty()) {
     const auto* nextUse = *uses_.begin();
     nextUse->user()->setInput(nextUse->inputIdx(), newInput);
   }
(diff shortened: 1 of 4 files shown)


flashlight/fl/tensor/backend/jit/ir/Node.cpp
```diff
@@ -77,13 +77,11 @@ const UseList& Node::uses() const {
...
 
 void Node::replaceAllUsesWith(Node* newInput) {
-  refCount_++; // a trick to avoid freeing this node in last iteration
-  while (!uses_.empty()) { // each iteration updates uses
+  // each iteration updates links an existing user to newInput
+  while (!uses_.empty()) {
     const auto* nextUse = *uses_.begin();
     nextUse->user()->setInput(nextUse->inputIdx(), newInput);
   }
-  newInput->uses_.splice(newInput->uses_.end(), std::move(uses_));
```
StrongerXi, line 85: The while loop takes care of the `Use` transfer.

jacobkahn: 👍

jacobkahn (approved):

PR #1038, opened 2022-10-19
Implement jit optimizer extension

StrongerXi:
### Summary
Use `TensorExtension` to support backend-specific optimization passes for JIT.

### Test Plan (required)
CI

### DIFF
--- a/flashlight/fl/tensor/backend/jit/opt/JitOptimizerExtension.h
+++ b/flashlight/fl/tensor/backend/jit/opt/JitOptimizerExtension.h
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include "flashlight/fl/tensor/TensorExtension.h"
+#include "flashlight/fl/tensor/backend/jit/opt/Pass.h"
+
+namespace fl {
+
+/**
+ * Tensor Extension to enable backend-specific JIT graph optimization.
+ */
+class JitOptimizerExtension : public TensorExtension<JitOptimizerExtension> {
+ public:
+  static constexpr auto extensionType = TensorExtensionType::JitOptimizer;
+
+  /**
+   * Get backend-specific optimization passes.
+   */
+  virtual std::vector<std::unique_ptr<Pass>> passes() = 0;
+};
+
+} // namespace fl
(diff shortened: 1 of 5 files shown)


flashlight/fl/tensor/backend/jit/opt/JitOptimizerExtension.h
```diff
@@ -0,0 +1,28 @@
...
+ * Tensor Extension to enable backend-specific JIT graph optimization.
+ */
+class JitOptimizerExtension : public TensorExtension<JitOptimizerExtension> {
+ public:
+  static constexpr auto extensionType = TensorExtensionType::JitOptimizer;
+
+  /**
+   * Get backend-specific optimization passes.
+   */
+  virtual std::vector<std::unique_ptr<Pass>> passes() = 0;
```
jacobkahn, line 25: I see this doesn't return a reference -- how does this work with optimizer lifetime? If I call `.passes()`, we have to move each `Pass` out of optimizer extension ownership? What happens to the obj after that?

flashlight/fl/tensor/backend/jit/opt/Optimizer.cpp
```diff
@@ -7,13 +7,34 @@
...
 
-Optimizer::Optimizer() {
-  // TODO figure out a configuration API (e.g., LLVM pass style macro)
+namespace {
+
+template <typename T>
+void extend(std::vector<T>& extendee, std::vector<T>&& elems) {
+  extendee.insert(
+      std::end(extendee),
+      std::make_move_iterator(std::begin(elems)),
```
jacobkahn, line 22: `#include <iterator>`

jacobkahn (changes requested):

flashlight/fl/tensor/backend/jit/opt/JitOptimizerExtension.h
StrongerXi: Yeah so the `passes()` here is really meant to be a pass constructor (makes new instances of the same passes upon every invocation, see #1040). Reasons are 2 fold:
1. To avoid sharing `Pass` -- right now it doesn't make a difference, but I'm thinking we might add state into `Pass` (for workload-specific optimization, like caching). In that case, An optimizer's passes should have their own state, for that optimizer's workload.
2. Going from shared `Pass` to a new copy is a bit more difficult -- we need a `Pass copy() virtual = 0` method on `Pass` .

StrongerXi:
Rebase and address feedback.

StrongerXi:
Closing because #1039 already incorporated the changes in this PR.

PR #1040, opened 2022-10-19
Implement onednn binop fusion pass

StrongerXi:
### Summary
Leverage OneDNN's post-ops via backend-specific JIT optimization pass!

The pass itself is heavily documented, and the tests should also illustrate _what_ the pass does.

This PR focuses on a simple & extensible proof-of-concept -- there are a lot more we can do with this pass, as documented in the code.

### Test Plan (required)
Unit tests are added
```
make JitOneDnnOpFusionTest
```


flashlight/fl/tensor/backend/jit/opt/backends/onednn/OneDnnOpFusion.h
```diff
@@ -0,0 +1,72 @@
...
+ * | n1 = n1 + n2 |
+ * | n1 = n2 + n3 |
+ * ----------------
+ *
+ * TODO
+ * - leverage commutativity of certain binops to bypass the rhs-only limitation
+ *   of OneDNN binary post-ops.
+ * - support more than just binop fusion
+ */
+class OneDnnOpFusion : public Pass {
```
jacobkahn, line 49: `OneDnnOpFusion` is quite generic -- is the idea here that this spans a large set of optimizations beyond binop fusion?

flashlight/fl/tensor/backend/jit/opt/backends/onednn/OneDnnOpFusion.cpp
```diff
@@ -0,0 +1,240 @@
...
+  // accumulatedBinopInfos: { { binop2->op(), x2 }, { binop1->op(), x1 } }
+  std::vector<BinopInfo> accumulatedBinopInfos;
+};
+
+namespace {
+
+// https://github.com/oneapi-src/oneDNN/blob/adeda9fcc20149effb1bffc051262810e9f3138c/include/oneapi/dnnl/dnnl_types.h#L2914
+static constexpr unsigned kOneDnnMaxNumPostOps = 32;
+
+dnnl::memory::data_type getOneDnnTypeWithLargestRange(
```
jacobkahn, line 44: nit -- might be good to detail this type promotion behavior somewhere

flashlight/fl/tensor/backend/jit/opt/backends/onednn/OneDnnOpFusion.cpp
```diff
@@ -0,0 +1,240 @@
...
+  // Assume `searchRoot == binop2`
+  //
+  // x0  x1
+  //  \  /
+  //  binop1  x2
+  //     \  /
+  //    binop2
+  //
+  // accumulatedBinopInfos: { { binop2->op(), x2 }, { binop1->op(), x1 } }
+  std::vector<BinopInfo> accumulatedBinopInfos;
```
jacobkahn, line 36: Why not use a `std::array<2>` or the like if we know there are only going to be two children to binops?

I suppose each fusion pass will have its own independent abstractions on search; makes sense to me.

flashlight/fl/tensor/backend/jit/opt/backends/onednn/OneDnnOpFusion.cpp
```diff
@@ -0,0 +1,240 @@
...
+  std::vector<Node*> inputNodes{node};
+  std::vector<dnnl::algorithm> algs;
+  for (int i = state.accumulatedBinopInfos.size() - 1; i >= 0; i--) {
+    const auto& info = state.accumulatedBinopInfos[i];
+    algs.push_back(binopToOneDnnAlg(info.op));
+    inputNodes.push_back(info.rhsNode);
+  }
+
+  // TODO refactor with common logic in OneDnnBackend
+  auto evalFunc = [algs = std::move(algs)](
```
jacobkahn, line 171: Great to have all of the goo we want to abstract away eventually in one place, nice.

jacobkahn (approved):
This looks great for now! Mainly a few questions, but I'm ready to go on this.

flashlight/fl/tensor/backend/jit/opt/backends/onednn/OneDnnOpFusion.h
StrongerXi: That's the intention indeed.

The thing with OneDNN's post-ops is that, we can append different ops in the chain (eltwise, sum, and even convolution). So ideally we don't wanna restrict the fusion to be "binops only".

I'll look more into how to cleanly do this after I throw in other IR nodes (eltwise, etc.)

flashlight/fl/tensor/backend/jit/opt/backends/onednn/OneDnnOpFusion.cpp
StrongerXi: Just added a documentation at the use site.

Eventually this will be replaced by a type inference mechanism in JIT.

flashlight/fl/tensor/backend/jit/opt/backends/onednn/OneDnnOpFusion.cpp
StrongerXi: > Why not use a std::array<2> or the like if we know there are only going to be two children to binops?

(If I understood correctly, you meant for `std::vector<BinopInfo>`? Because `BinopInfo` doesn't hold homogeneous fields.). The diagram shows only an example scenario. We could have arbitrary # of accumulated binops here (welp max 32 per OneDNN).

> will this structure eventually hold a vector<OpInfo> or something of the like?

That's what I have in mind, need to explore a bit more though. As explained in a comment above, ideally we need to support fuse heterogeneous ops in a single pass (to fully leverage OneDNN postops).

StrongerXi:
Rebase & address all feedbacks.

PR #1047, opened 2022-10-28
Expose non R-value releaseAdapter

bwasti:
Related to https://github.com/facebookresearch/shumai/issues/50

### Summary

Allows the deletion of all underlying adapter values without destroying the Tensor object itself.


jacobkahn (approved):
💀

PR #1049, opened 2022-11-01
Implement copy and in place assignment

StrongerXi:
### Summary
Use SSA to implement assignment in the lazy JitTensor.

This allows us to capture maximal graph that obeys the computation result of tensors under assignment, based on the lexical order of their definitions.

Lazy computation and assignment doesn't play well with each other. The alternative I considered was in-place node mutation (so assignment corresponds to changing graph), and aggressively evaluate subgraphs that uses the to-be-updated node, because they rely on the "old value". That approaches is complicated, and fails to capture maximal graph.

Granted, SSA IR seems to deprive "buffer reuse" during assignment. But we could add backend evaluator/optimizer logic to recover this; moreover, allowing the backend to make global decision on such buffer reuse could be more optimal (than having programmer making such decision).

### Test Plan
Unit tests are added

```
make JitTensorTest
```

### DIFF
--- a/flashlight/fl/tensor/backend/jit/JitTensor.h
+++ b/flashlight/fl/tensor/backend/jit/JitTensor.h
@@ -20,8 +20,8 @@ namespace fl {
 template <typename T>
 class JitTensor : public JitTensorBase {
  protected:
-  Tensor fromNode(Node* node) const override {
-    return toTensor<JitTensor>(node);
+  Tensor fromSharedNode(std::shared_ptr<SharedData> sharedNode) const override {
+    return toTensor<JitTensor>(std::move(sharedNode));
   }
 
   TensorBackend& wrappedBackend() const override {
@@ -50,6 +50,8 @@ class JitTensor : public JitTensorBase {
   }
 
   explicit JitTensor(Node* node) : JitTensorBase(node) {}
+  explicit JitTensor(std::shared_ptr<SharedData> sharedNode)
+      : JitTensorBase(std::move(sharedNode)) {}
 
   // TODO SPoC for these defaults (also in OneDNN backend)
   JitTensor() : JitTensor({0}, fl::dtype::f32, nullptr, Location::Host) {}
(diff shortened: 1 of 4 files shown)


flashlight/fl/tensor/backend/jit/JitTensor.h
```diff
@@ -20,8 +20,8 @@ namespace fl {
 template <typename T>
 class JitTensor : public JitTensorBase {
  protected:
-  Tensor fromNode(Node* node) const override {
-    return toTensor<JitTensor>(node);
+  Tensor fromSharedNode(std::shared_ptr<ShareData> sharedNode) const override {
```
jacobkahn, line 23: To make sure I understand correctly — a shared ptr to `SharedData` serves only to pass refcount operations themselves on `Node`s around?

jacobkahn (approved):
LGTM -- happy to merge, cc @bwasti if you have any thoughts

flashlight/fl/tensor/backend/jit/JitTensor.h
StrongerXi: > SharedData serves only to pass refcount operations themselves on Nodes around

Actually no. `RefCount` and `SharedData` solves orthogonal problems -- the former is for memory management of `Node`, the latter is for mutation under aliasing (alias from shallowCopy or indexing).

Note that we used `SharedData` to handle alias at many other places -- [Variable::SharedData](https://github.com/flashlight/flashlight/blob/main/flashlight/fl/autograd/Variable.h#L327-L328), [ArrayFireTensor](https://github.com/flashlight/flashlight/blob/main/flashlight/fl/tensor/backend/af/ArrayFireTensor.h#L30-L32), and [OneDnnTensor](https://github.com/flashlight/flashlight/blob/main/flashlight/fl/tensor/backend/onednn/OneDnnTensor.h#L60-L61).

StrongerXi:
Rebase & address feedbacks.

PR #1051, opened 2022-11-01
Implement simple shape inference

StrongerXi:
Depends on #1049 

### Summary

Populate shape information during construction of JIT graph nodes.

This avoids forcing evaluation if user calls `tensor.shape()` (maybe they don't care about the tensor result at all).

### Test Plan (required)
Unit tests are added

```
make JitNodeTest
```

### DIFF
--- a/flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
+++ b/flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
@@ -64,7 +64,7 @@ TensorBackendType JitTensorBase::backendType() const {
 }
 
 const Shape& JitTensorBase::shape() {
-  FL_JIT_TENSOR_UNIMPLEMENTED;
+  return node()->shape();
 }
 
 fl::dtype JitTensorBase::type() {
--- a/flashlight/fl/tensor/backend/jit/JitBackend.cpp
+++ b/flashlight/fl/tensor/backend/jit/JitBackend.cpp
@@ -341,10 +341,6 @@ Tensor JitBackend::argsort(
   }                                                                           \
   FL_JIT_BINARY_OP_LITERALS_DEF_STUB(FUNC, OP);
 
-FL_JIT_BINARY_OP_LITERALS_DEF_STUB(add, +);
-FL_JIT_BINARY_OP_LITERALS_DEF_STUB(sub, -);
-FL_JIT_BINARY_OP_LITERALS_DEF_STUB(mul, *);
-FL_JIT_BINARY_OP_LITERALS_DEF_STUB(div, /);
 FL_JIT_BINARY_OP_DEF_STUB(==, eq);
 FL_JIT_BINARY_OP_DEF_STUB(!=, neq);
 FL_JIT_BINARY_OP_DEF_STUB(<, lessThan);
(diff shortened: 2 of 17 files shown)


flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
```diff
@@ -15,44 +15,56 @@
 
 namespace fl {
 
-JitTensorBase::JitTensorBase(Node* node) : node_(node) {
-  node_->incRefCount();
-}
+struct JitTensorBase::ShareData {
```
bwasti, line 18: doesn't `shared_ptr<Node>` handle this kind of stuff for you? curious why you're wrapping it twice

flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
```diff
@@ -15,44 +15,56 @@
...
-// take care of refcount for old & new nodes
 void JitTensorBase::replaceNode(Node* newNode) const {
-  if (node_ != newNode) {
-    node_->decRefCount();
+  auto oldNode = node();
+  if (newNode != oldNode) {
     newNode->incRefCount();
-    const_cast<JitTensorBase*>(this)->node_ = newNode;
+    oldNode->decRefCount();
+    sharedNode_->node = newNode;
```
bwasti, line 41: if you did use shared_ptr, some built-in compiler optimizations can actually avoid a refcount change here
```cpp
replaceNode(std::shared_ptr<Node>& newNode) const {
  node_.swap(newNode);
}
```

flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
```diff
@@ -15,44 +15,56 @@
...
+JitTensorBase::JitTensorBase(Node* node)
+    : JitTensorBase(std::make_shared<ShareData>(node)) {}
+
+JitTensorBase::JitTensorBase(std::shared_ptr<ShareData> sharedNode)
+    : sharedNode_(sharedNode) {}
+
+JitTensorBase::~JitTensorBase() {}
 
-// take care of refcount for old & new nodes
 void JitTensorBase::replaceNode(Node* newNode) const {
```
bwasti, line 36: is this const correct?

flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
jacobkahn: Bleh -- is having `weak_ptr` actually that much worse if you're just dealing with `Node*` in the public API? It seems like the `ShareData` is just internal sugar to pass around refcount ops and users will still just want `Node*` to build their own passes

Also might be worth thinking about the level of technical-ness one could expect from someone implementing compiler passes 😄 

Probably should've put this thought on your prior PR.

StrongerXi:
Rebase.

flashlight/fl/tensor/backend/jit/JitTensorBase.cpp
StrongerXi: Had some discussions offline. Current plan (for a future refactor PR, after landing the remaining stack which will be orthogonal to this):

1. Use `std::shared_ptr<Node>` to replace manual refcounting.
2. Use `Node&` or `std::weak_ptr<Node>` inside `Use` (need to investigate more)
3. Use something other than refcount to determine whether to retain tensor result in Evaluator.

PR #1052, opened 2022-11-02
Implement basic indexing (w/o indexed update or multiple tensor indices)

StrongerXi:
### Summary

Tldr as title.

A few key points, otherwise the documentation in the code offers more explanations
- Indexing essentially creates a special shallowCopy to the indexed tensor that looks only at the indexed region.
- since we handle assignment with SSA, we must re-materialize the indexed node whenever the underlying data changes.

### Test Plan (required)
New unit tests are added
```
make JitNodeTest
make JitEvaluatorTest
make JitTensorTest
```

PR #1055, opened 2022-11-16
Implement indexed update

StrongerXi:
Depends on #1052 

### Summary
Support indexed update while retaining maximal graph and preserving computation semantics. This requires a new `IndexUpdateNode` because we the "indexed update" semantics cannot be composed from existing nodes.

Example:

```c++
Tensor t = ...;
Tensor add = t + t;
Tensor t0 = t(0);
t(0, 0) = 42;
// 1. both t and t0 are affected (if one prints them out, update is reflected)
// 2. `add` isn't affected (old graph remains intact -- graph computation semantics is immutable)
// 3. doesn't matter which tensor we use to trigger evaluation (lazy evaluation remains transparent to user)
```

### Test Plan (required)
New unit tests are added
```
make JitNodeTest
make JitEvaluatorTest
make JitTensorTest
```


StrongerXi:
Fix build failure by removing unnecessary includes.

StrongerXi:
Rebase.

StrongerXi:
Rebase.

flashlight/fl/tensor/backend/jit/eval/Evaluator.cpp
```diff
@@ -52,16 +52,45 @@ void Evaluator::evalCustomNode(CustomNode& node) {
...
 
 void Evaluator::evalIndexNode(IndexNode& node) {
-  std::vector<Index> indices;
-  for (const auto& index : node.indices()) {
+  const auto& indexedTensor = node.indexedNode()->getResult().value();
+  node.setResult(indexedTensor(unwrapTensorInIndices(node.indices())));
+}
+
+void Evaluator::evalIndexedUpdateNode(IndexedUpdateNode& node) {
+  // TODO no need to copy if indexedNode has only 1 user here
```
jacobkahn, line 60: Why not implement it? Seems like a very useful optimization. Feel free to add in a downstream PR too

flashlight/fl/tensor/backend/jit/eval/Evaluator.cpp
```diff
@@ -52,16 +52,45 @@ void Evaluator::evalCustomNode(CustomNode& node) {
...
+  auto indexedTensor = node.indexedNode()->getResult().value().copy();
+  const auto firstUnwrappedIndices =
+      unwrapTensorInIndices(node.indexings().front());
+  const auto& updateDataTensor = node.updateDataNode()->getResult().value();
+  // if we do X = Y, it's a copy instead of update, thus the special case here
```
jacobkahn, line 65: I think I'm not understanding the comment -- why would assignment result in an IndexedUpdateNode?

flashlight/fl/tensor/backend/jit/ir/IndexedUpdateNode.h
```diff
@@ -0,0 +1,54 @@
...
+ * we have
+ *   output == indexedTensor __after__
+ *      indexedTensor(indexings[0])(indexings[1])... = updateDataTensor
+ *
+ * But note that since the graph represents pure computation (SSA, no
+ * assignment), this node conceptually creates a new tensor that holds the
+ * result of the indexed update, thus the name.
+ */
+class IndexedUpdateNode : public NodeTrait<IndexedUpdateNode> {
+  const std::vector<std::vector<Index>> indexings_;
```
jacobkahn, line 29: So understand correctly -- since this is literally `fl::Index` rather than some symbolic type; the Tensor represented by the index here might be (should be) a JIT tensor, so the rest of the interop works since that itself corresponds to some node?

jacobkahn (approved):
Nice -- a few suggestions/questions. Good to go though.

PR #1065, opened 2022-12-03
Cover all ops in JIT frontend

StrongerXi:
### Summary
As title. This enables 
1. applying JIT to _almost_ any existing application
2. capturing maximal graph (which will become immensely helpful with the upcoming GraphVizPrinter & performance profiling).

The changes are quite simple -- just apply `CustomNode` to inject evaluation logic for all the remaining ops, and shape inference for certain ops.

There are small TODOs I dropped and minor changes I plan to send out in a few upcoming PRs, but they are pretty niche, and shouldn't affect overall applicability of the JIT.

### Test Plan (required)

I manually used `fl::setDefaultTensorType<JitTensor<DefaultTensorType_t>>` in all existing tensor tests, e.g., `TensorBaseTest` to make sure things are correct (except for about 5% due to the TODOs I mentioned above).

### DIFF
--- a/flashlight/fl/tensor/backend/jit/JitBackend.cpp
+++ b/flashlight/fl/tensor/backend/jit/JitBackend.cpp
@@ -8,17 +8,51 @@
 #include "flashlight/fl/tensor/backend/jit/JitBackend.h"
 
 #include <cassert>
+#include <iostream>
+#include <sstream>
 #include <stdexcept>
 
 #include "flashlight/fl/tensor/TensorBase.h"
 #include "flashlight/fl/tensor/backend/jit/JitTensorBase.h"
+#include "flashlight/fl/tensor/backend/jit/ShapeInference.h"
 #include "flashlight/fl/tensor/backend/jit/ir/ScalarNode.h"
-
-#define FL_JIT_BACKEND_UNIMPLEMENTED \
-  throw std::invalid_argument(       \
-      "JitBackend::" + std::string(__func__) + " - unimplemented.");
+#include "flashlight/fl/tensor/backend/jit/ir/ValueNode.h"
 
 namespace fl {
+
+namespace {
+
+std::vector<Node*> tensorsToNodes(const std::vector<Tensor>& tensors) {
+  std::vector<Node*> nodes;
+  // JitTensor copy is ~free, but whatever
+  for (const auto& tensor : tensors) {
+    nodes.push_back(toJitTensorBase(tensor).node());
+  }
+  return nodes;
+}
+
+template <typename... T>
+std::vector<Node*> tensorsToNodes(const T&... tensors) {
+  std::vector<Node*> nodes;
+  // JitTensor copy is ~free, but whatever
+  for (const auto& tensor : {&tensors...}) {
+    nodes.push_back(toJitTensorBase(*tensor).node());
+  }
+  return nodes;
+}
+
+template <>
+std::vector<Node*> tensorsToNodes() {
+  return {};
+}
+
+const Tensor& materialize(Tensor tensor) {
+  auto& jitTensor = toJitTensorBase(tensor);
+  jitTensor.eval();
+  return jitTensor.node()->getResult().value();
+}
+
+} // namespace
 
 JitBackend::JitBackend(
     TensorBackend& wrappedBackend,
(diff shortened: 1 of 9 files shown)


flashlight/fl/tensor/backend/jit/JitBackend.cpp
```diff
@@ -111,253 +351,303 @@ Tensor JitBackend::fullWithType(const Shape& shape, T value, dtype type) {
...
+  return jitTensorCreator_(CustomNode::create(
+      "reshape",
+      tensorsToNodes(tensor),
+      Shape(shape),
+      [this, shape](auto inputs) {
+        return wrappedBackend_.reshape(*inputs.at(0), shape);
+      }));
+}
+
+Tensor JitBackend::transpose(const Tensor& tensor, const Shape& axes = {}) {
```
jacobkahn, line 393: Any plans to not use CustomNodes for these ops?

Possible my understanding of the existing `CustomNode` abstraction is wrong, but it seems like using a `CustonNode` for transpose precludes building generic passes that might affect it -- for example, removing an idempotent double transpose (`transpose(transpose(t))`. Possible motivation for a `UnaryOp`?

jacobkahn (commented):
Some preliminary comments. Going to return to reviewing this after it's rebased on changes that aren't yet merged.

flashlight/fl/tensor/backend/jit/JitBackend.cpp
StrongerXi: > Any plans to not use CustomNodes for these ops?

Totally, I should've mentioned this.

Right now `CustumNode` serves 2 purposes:
1. Inject backend-specific optimization (OneDNN post-op fusion)
2. Cover ops we currently don't have optimization support for (i.e., `CustomNode` is opaque to optimization passes).

My goal is to keep IR minimal by following this workflow:
1. Identify hotspot via the upcoming GraphVizPrinter in workload we care about
2. Come up with optimization passes
3. Augment IR if needed, e.g., `TransposeNode` if we want to support `transepose(transpose(t)) --> t`

I want to make sure we don't add IR nodes & optimization passes for the sake of adding them -- they should be frequent or significant enough in the workload we wish to optimize.

flashlight/fl/tensor/backend/jit/JitBackend.cpp
```diff
@@ -8,19 +8,253 @@
...
+    throw std::runtime_error("[inferPadOutputShape] too many paddings");
+  }
+  for (unsigned axis = 0; axis < padWidths.size(); axis++) {
+    const auto& [beforeDim, afterDim] = padWidths[axis];
+    outputDims[axis] += beforeDim + afterDim;
+  }
+  return Shape(outputDims);
+}
+
+Shape inferMatmulOutputShape(
```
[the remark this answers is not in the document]
StrongerXi: Moved them to a `ShapeInference.h` file, and dropped a TODO suggesting we refactor it with AF & OneDNN backend, and allow future backend implementation to access and reuse them.

Regarding looking up shape inference,

1. I'll likely add a `matmul` node since it behaves a bit different from regular binops (batching and transpose).
2. Right now I'm attaching the logic to `Node` subclasses (e.g., inside `BinaryNode::create`). I do know that existing frameworks like TVM has a decentralized way of storing such logic (shape, type inference). But I kinda wanted to avoid having another `ShapeInference` pass that annotates `Node::optionalShape`, basically KISS for now I guess, and see how things evolve:).

StrongerXi:
Rebase & address feedbacks.

StrongerXi:
For binop with scalar input, convert scalar to a tensor of shape `<1, 1, ..., 1>` instead of matching the shape of the other tensor input. This drastically reduces runtime and memory usage in multiple places.

flashlight/fl/tensor/backend/jit/ShapeInference.cpp
```diff
@@ -0,0 +1,213 @@
...
+    const Shape& rhsShape,
+    MatrixProperty lhsProp,
+    MatrixProperty rhsProp) {
+  std::vector<Dim> lhsDims = lhsShape.get();
+  std::vector<Dim> rhsDims = rhsShape.get();
+  const auto lhsRank = lhsDims.size();
+  const auto rhsRank = rhsDims.size();
+  const bool isLhsScalarOrVector = lhsRank <= 1;
+  const bool isRhsScalarOrVector = rhsRank <= 1;
+  if (isLhsScalarOrVector) { // pad to (1 x 1/K)
```
jacobkahn, line 172: Why does this help with oneDNN? Fine with it, but worth noting that there are many cases in which frameworks will do this broadcast implicitly, and where it might actually be useful to distinguish between a scalar (which is typically 0D shape-wise) and a 1-dimensional tensor.

PR #1094, opened 2023-03-21
[trivial] Fix macOS build

jacobkahn:
Apple LLVM is stricter about headers.

Test plan: local build, CI


jacobkahn:
fixed in https://github.com/flashlight/flashlight/pull/1096/files

PR #1116, opened 2023-05-26
Switch to setup-micromamba

pavelzw:
`mamba-org/provision-with-micromamba` is deprecated.


jacobkahn:
Thanks for the fix @pavelzw

PR #1130, opened 2023-06-12
CI baseline to build docs

jacobkahn:
See title. Build via doxygen, sphinx, and breathe.

Test plan: CI


jacobkahn:
Action is only on PR open, not re-open.

PR #1158, opened 2023-10-28
Add templated forward method to base Module class to reduce boilerplate

benborder:
### Summary
The forward interface for `Module` is a bit limiting and leads to a bit more boiler-plate code than is really necessary. This adds a templated variadic forward method to `Module` which dynamically dispatches the correct forward method for `UnaryModule` and `BinaryModule` derived types or otherwise uses the standard forward method of Module. This has the following advantages:
- Reduces boiler-plate code particularly when working with a type erased module.
- For unary/binary based modules (and in particular the `Sequential` container) this avoids allocating a vector for each forward method called.
- The current interface remains unchanged, so existing code will not be affected.
- Is restricted via SFINAE to only allow the type `Variable`as input arguments.

However this approach does assume that if a single parameter is parsed in, then only a single parameter is output. I think this is a reasonable assumption as the current vector based wrapper forward method can be used to avoid this issue.

The primary change is in `Module.h`, the remaining changes are taking advantage of the new forward method. For reference I used the following regex search in VSCode to help locate potential changes. find:
- `forward\(\{([^,}]+)\}\)\.front\(\)`
- `forward\(\{([^,}]+)\}\)\[0\]`
- `forward\(\{([^}]+)\}\)`

replace: `forward($1)`

### Test Plan (required)
Use CI to ensure builds pass.


benborder:
@jacobkahn and @r-barnes Seems the CI is failing due to https://github.com/flashlight/flashlight/commit/04e72e4e18287128ace65783d5aedd0bd6f0a3d9

benborder:
For some background. I had considered an alternative approach to remove the Unary and Binary Module interfaces and add a unary forward method to `Module` and instead of making them pure virtual to provide an interface definition, simply have a default virtual implementation which throws an unimplemented exception.

This approach is a bit simpler (no templating tricks etc), but it loses the ability to enforce a specific interface is implemented at compile time. This means things like `Sequential` must check the result of each forward pass has the correct number of elements at runtime. Ultimately it would make the API more complex due to a less well defined and less strict API.

